PR #14, opened 2015-05-27
Update fr_FR.lang

Mazdallier:
this is the turn off all small ans tiny dust, ingot and nugget, ore crushed, rod and plate


modmuss50:
Thanks again

Issue #19, opened 2015-06-02
Thermal Generator UI does not update when fluid is pumped in


joflashstudios:
When lava is pumped into the thermal generator with Thermal Dynaics fluiduct instead of manully placed via buckets, the thermal generator UI reports it is empty, but it still generates EU.

joflashstudios:
Quitting and re-opening the world causes the contents to appear.

PR #20, opened 2015-06-05
Added Semifluid Generator

joflashstudios:
Added functionality to the semifluid generator block. Currently supports creosote, biomass, oil, fluidsodium, and fluidlithium, with the EU numbers from http://ftbwiki.org/Semifluid_Generator_%28GregTech%29. Suffers from issue #19, I haven't been able to figure that out. Adding fluid manually when #19 is in effect results in more strange behavior, especially if you add a different supported fluid than the one that was already there, but hidden.
There are a number of config todos related to this machine.


modmuss50:
Great thanks, I will fix #19 soon.

PR #23, opened 2015-06-05
Added gas turbine. Still missing recipe and some config elements.

joflashstudios:


joflashstudios:
Also, I'm having trouble finding where recipes for crafting actual machines are defined. @modmuss50 , can you point me in the right direction?

modmuss50:
If you want to add a recipe to current machines look in techreborn.init.ModRecipes. On line 171 we add the recipes to current machines. If you want to hook into the handler just look and see how the other ones are done. If you like tell me how many inputs/outputs you want and I will add it.

PR #68, opened 2015-06-29
a couple fixes

TheQuantumGeneral:
only 3 edits, it just makes the thermal generator only accept fluid containers on the input and fixing the chlorine crash from the electrolyzed salt recipe.


modmuss50:
Looks good :) thanks

PR #109, opened 2015-08-13
Finished Cloaking Device

primetoxinz:
as per title


modmuss50:
Need to think of a way to fix potions.

PR #131, opened 2015-08-23
Update zh_CN.lang

Amamiya-Nagisa:


modmuss50:
Thanks for that, I will merge that when I get back from holiday in a few days.

PR #200, opened 2015-10-22
Fixed #199

### LINKED ISSUE #199: Remain In motion
Remain in motion bug in tooltip (log crach)

http://hastebin.com/upigulosiy.avrasm

MrBretze:


modmuss50:
I've done it my way because then it will only work on our blocks and items, there could be other mods that also crash by creating a fake tile

PR #346, opened 2016-02-21
Update some textures, fix some sounds, add scrap texture

ProfessorProspector:


modmuss50:
Ive merged some of your commits, the one with the rubber tree changes would break them.

modmuss50:
Thanks for that! I will have to close this, but I have merged most of this PR

ProfessorProspector:
Why would it break them? Seemed to work fine in my env

ProfessorProspector:
Hmm it appears for some reason the java files didn't get pushed to git. I'll have to figure that out in the morning.

modmuss50:
That would explain it

ProfessorProspector:
Anyway the reason I changed it in the first place was because the tinker part builder was treating the top face as the side face because of the texture name. Place a rubber wood part builder down and see for yourself.

PR #624, opened 2016-06-05
Textures for andesiteDust, graniteDust, and dioriteDust

brnbrd:


Prospector:
These have already been replaced for the retexture, but until then, this works. Thanks ;)

brnbrd:
You're welcome. How is the progress on said retexture?

Prospector:
Mostly done, just need to do some machines and random things here and there

PR #628, opened 2016-06-05
Added grinder recipes for gems -> dust

brnbrd:


modmuss50:
Nice, thanks again 👍

brnbrd:
modmuss50, do you know where I could get a link to GregTech 2.78 for testing?

Prospector:
are you on our IRC channel? come on and I'll give you a download #TechReborn on esper

brnbrd:
Just joined

Yulife:
We probably need to do one big commit where we mirror all recipes from the old days, right now most machines recipes are not done

modmuss50:
They do not need to be exact, as long as everything fits in we should be good.

PR #638, opened 2016-06-07
GUI Sheet 

gigabit101:


Yulife:
ew why are there pixels behind the icons

Yulife:
I'll fix that

PR #670, opened 2016-07-14
Fix CraftTweaker compatibility

### LINKED ISSUE #653: Handler mods.techreborn doesn't work, not sure what else it could be.
Tried to make a recipe in the Rolling Machine if that makes any difference. But it specifically stated "No such member: techreborn".

ghost:
A version of CraftTweaker that is compatible with 1.9.4 has been
released since the Minetweaker/CraftTweaker compatibility was removed a while back. I
have restored compatibility with the new version of CraftTweaker and it seems to work like a charm.

Fixes #653


Prospector:
hopefully it works lol

PR #671, opened 2016-07-15
Properly fluid rendering in dynamic cells. Temporary fix of recipes incapability…

Dragon2488:
…with renamed (or tagged) cells. Merge dynamic and empty cells into one item. Right click on fluid block fills cell with fluid.


modmuss50:
Thanks, i will be back working on tr in two weeks after my holiday.

Dragon2488:
It actually works!
http://i.imgur.com/7ameIsG.png
(left - IC2 cell, middle - our cell, right - normal item)

PR #672, opened 2016-07-15
Properly integration with IC2 through ore dictionary. JEI recipes...

Dragon2488:
show energy usage, time and some additional recipes properties like heat capacity.


Prospector:
wow you're on a roll

modmuss50:
Oh wow, amazing work! 👍

Dragon2488:
Thanks.

Dragon2488:
@ProfessorProspector I realized that it isn't latest version and code should be written again for 1.10.2, isn't it? If it is, sad :(

ghost:
@Dragon2488 I'm pretty sure that this is the main branch because it is cross compatible with 1.10.2 By the way, I was just testing out the new jar file and it looks like the game will crash on startup if IC2 isn't installed because it can't reference IC2 classes.

Relevant error log lines:
`net.minecraftforge.fml.common.LoaderException: java.lang.NoClassDefFoundError: ic2/core/item/type/CraftingItemType
.
.
.
.
Caused by: java.lang.NoClassDefFoundError: ic2/core/item/type/CraftingItemType
    at techreborn.Core.init(Core.java:128)`

Dragon2488:
@al132 Can you provide full log please? I can't see any mentions of IC2 outside Loader.isModLoaded("IC2").

ghost:
Here it is. Now i'm very new to modding so I could be mistaken, but I think that the problem is that you are using a reference from what I presume is the dev build (ic2/core/item/type...) when we only have the API packaged with the mod, which would have a different way to refer to IC2 items.
[fml-client-2.txt](https://github.com/TechReborn/TechReborn/files/367151/fml-client-2.txt)

ghost:
Now that I look further, perhaps none of it is packaged with the JAR. I really don't know my way around the build file at all, but perhaps this could be the problem https://github.com/TechReborn/TechReborn/blob/4fcefc84f31b1410acd323de6dd845939afcfced/build.gradle

```

jar {
    exclude "**/*.psd"
    classifier = 'universal'
//    configurations.shade.each { dep ->
//        from(project.zipTree(dep)){
//            include 'ic2/api/'
//            include 'forestry/api/fuels/'
//            exclude 'META-INF', 'META-INF/**', '**/*.java'
//        }
//    }
}
```

Dragon2488:
API should be provided by IC2, not packaged in the jar. However it's strange that integration code passes IC2 check through Loader.isModLoaded when no IC2 installed.

modmuss50:
Yeah there should be no calls to core ic2 things. And it should work without it.

And the 1.9.4 branch is set to build 1.9.4 1.10 and 1.10.2 versions on the jenkins

Dragon2488:
Ok. Cool.

PR #673, opened 2016-07-16
Implemented Lightning Rod, improved ore dictionary integration…

Dragon2488:
…for dusts, nuggets and plates.  Plates now register automatically for every gem and ingot in TR.


Dragon2488:
And also question for @modmuss50: why RebornCore sources broken?

modmuss50:
I'm not sure, I will look at in a few when I get back to my pc

PR #679, opened 2016-07-18
Fixed Output of extractor

### LINKED ISSUE #658: Cannot automatically remove products of Extractor
1.9.4-1.3.2.120-universal Running in 1.10

Cannot automatically remove products of Extractor using any method.

ghost:
Fixes #658 

I looked off of how the grinder was implemented, which has proper input/output functionality and is essentially an identical machine except for the recipes. My guess is that @modmuss50 wants all of the machines to be implemented in a similar way, transitioning away from ISidedInventory although I may be mistaken. If this is the case I would be up for fixing more machines if it's just a matter of busy work that has been holding it back from being done.


modmuss50:
Yeah, this is correct, I near got around to moving the rest of the machines to the new system.

ghost:
Would you like all of the machines to be done this way? Or are there some that should be specifically sided? Like the furnace?

modmuss50:
No point doing it on blocks that all ready work for now. I have forgotten which ones it needs to be done on.

PR #686, opened 2016-07-21
Wires now transfer energy from all mods and manages it as EU.

Dragon2488:
Multipart will fire in hell.


modmuss50:
Check you emails for the invitation to push to the repo

Dragon2488:
Thanks.

PR #693, opened 2016-07-25
Fix Tinkers Construct crashing

PrincessOfEvil:
Probably.


modmuss50:
If this doesnt work, I have a fix in mind, poke me and I may be able to make it

PR #701, opened 2016-07-30
Fix typo in config file

Hobby-boy:
Client was spelt wrong


modmuss50:
I dont even think that config setting has been used since 1.7

Dragon2488:
Configs need cleanup

PR #721, opened 2016-08-06
Fixe lava cell dupe

wormzjl:


modmuss50:
I dont think this needs fixing in 1.10?

modmuss50:
I can check it at some point

Issue #484, opened 2016-03-29
Electrolyzer recipes


mystise:
Running TechReborn: 1.0.3.140 RebornCore: 2.0.4.44

1. Pyrite returns only 1 sulfur

![2016-03-29_15 42 51](https://cloud.githubusercontent.com/assets/246232/14124856/22e34148-f5c5-11e5-9beb-c680da93598f.png)

GT2: 
![2016-03-29_15 42 53](https://cloud.githubusercontent.com/assets/246232/14124860/26208c1c-f5c5-11e5-988c-eb246b2fcee3.png)

2. Sulfuric acid processing returns sulfur dust rather than sulfur cell (Not that bad since sulfuric acid is uncraftable (because the canning machine doesn't exist, so sulfur cells are uncraftable) and doesn't have any other uses)

![2016-03-29_15 45 47](https://cloud.githubusercontent.com/assets/246232/14124908/7e197dc0-f5c5-11e5-97aa-dfb48b2bfbdf.png)

GT2:
![2016-03-29_15 47 22](https://cloud.githubusercontent.com/assets/246232/14124920/8a578b54-f5c5-11e5-865f-3755c8c5b33e.png)

3. (Sapphire/peridot/pyrope dust)/sand return more empty cells than they consume

![2016-03-29_15 49 55](https://cloud.githubusercontent.com/assets/246232/14125016/27e77848-f5c6-11e5-9c49-0aac6612bb19.png)
![2016-03-29_15 53 29](https://cloud.githubusercontent.com/assets/246232/14125059/6c57438c-f5c6-11e5-95f5-d1a5049e0164.png)
![2016-03-29_15 56 12](https://cloud.githubusercontent.com/assets/246232/14125164/e52694fc-f5c6-11e5-8a69-804884bd3db2.png)
![2016-03-29_15 56 14](https://cloud.githubusercontent.com/assets/246232/14125165/e533d702-f5c6-11e5-912c-81a6af4371fd.png)

4. Recipes that return compressed air in GT2 should probably not take or return empty cells, as compressed air doesn't exist
- Ruby dust
- Sapphire dust
- Peridot dust
- Emerald dust
- Missing recipe placeholder (Flint dust?)
- Olivine (Currently Peridot, see 5) dust
- Pyrope dust
- Sand
- Almandine dust
- Spessartine dust
- Andradite dust
- Grossular dust
- Uvarovite dust
- Nitrogen Dioxide

[...]

Issue #10, opened 2015-05-24
Minor outstanding sync issue


Tntrololol:
Ghost items are still left behind when removing inputs from a machine.

Tntrololol:
Update on this: the problem seems to occur only when removing items when a machine is in the middle of trying to process something.

PR #728, opened 2016-08-10
Electrolyzer recipes

Zygimantass:
Reference to issue #484 
no.1 fixed
no.2 doesnt need fixing, as you said
no.3 fixed
no.4 don't understand this one
no.5 AFAIK Olivine dust doesn't have a recipe (correct me if I'm wrong)
no.6 fixed
no.7 fixed
no.8 nitrogen dioxide doesnt work?
no.9 couldnt fix
no.10 i don't think this needs to be fixed
no.11 same as #2
no.12 same as #10


modmuss50:
Great thanks

Zygimantass:
another pr incoming

PR #729, opened 2016-08-10
fixed #722

### LINKED ISSUE #722: Free saltpeter for everyone!
Hey guys!

We ran into a cute issue this morning as can be seen here: http://i.imgur.com/ExNQbia.gifv

When there are no items in the input slot of the grinder, it idly makes saltpeter.  Two at a time, it makes two more when you remove whatever is in the output slot.  I have had two separate users report this.

I removed all craft/modtweaker scripts to confirm this was not a script gone awry.  This happens in SSP and SMP.

I'm running FML 2046, TechReborn 1.4.0.41 and RebornCore 2.6.2.15 on Minecraft 1.10.2

Zygimantass:
Fixed #722 with a null ore check.

Fixed saltpeter, sulfur and maybe something else that doesnt have an ore.


modmuss50:
Thanks

PR #781, opened 2016-10-17
Fixed the blast Furnace recipe

DreiPixel:
does not use superconductors now. instead it uses the same recipe as the 1.9.4 branch which is cupronickel Heating coils instead of superconductors...


modmuss50:
Thanks for that, I must have missed merging this at some point when merging the branches. 👍

PR #815, opened 2016-11-28
Fixes to the Centrifuge Block

brad18:


src/main/java/techreborn/tiles/TileCentrifuge.java
```diff
@@ -28,29 +30,31 @@
 
 	public int tickTime;
 	public Inventory inventory = new Inventory(11, "TileCentrifuge", 64, this);
+	public UpgradeHandler upgradeHandler;
```
modmuss50, line 33: Im not sure this does anything atm, but it wont harm being their.

modmuss50:
Yeah, looks good. Thanks again :+1:

PR #831, opened 2016-12-13
Make treetap working with the IC2 rubber tree (Fix #786)

### LINKED ISSUE #786: [Suggestion] Treetap should consider IC2 rubber spots valid
I thought rubber is rubber whatever we call it but it seems like not. I love your treetap texture (it's just more fancier than the really old and rusty IC2 one) and I really hoped I could use it on my IC2 rubber trees but it does not work. Could you please make it work against IC2 trees? I really don't care what it produces (either IC2 resin or TR counterpart), but please, please make it work on IC2 rubber spots. Because your treetap is just better.

Modpack: FTB Presents Hermitcraft 1.0.1
Mods: IC2 2.6.69-ext110 & TechReborn 1.10.2-1.4.6.67

marysaka:
Changes:
- The dependency of IC2 have been updated to the last version
- This implementation follow the original treetap logic but use
TechReborn original electrical cost.


modmuss50:
Looks great, thanks.

PR #834, opened 2016-12-14
Add fluidstack info to jei recipe for IndustrialGrinder

Ourten:
This PR add the **fluid needed for a craft** to the IndustrialGrinder JEI recipes.

Beside the _IndustrialSawmill_, for which i haven't found **any recipes** other than minetweakers, there is no other machines inputting or ouputting fluidstacks in the mod according to my search. This is why i've made this in the form of a **specific code**.

The method _IndustrialGrinderRecipeWrapper#getFluidInputs_ is not linked to the ingredients, so i've considered wiser to not disturb the internal mechanics of the _BaseRecipeWrapper_ to make a **generic solution**.


modmuss50:
Looks good, thanks again.

PR #837, opened 2016-12-14
Energy generators fix

Ourten:
So i've encountered some bugs while using the fluid-based generators.

This PR will fix some of them : 

- It correctly update the **blockstate** of the machine when it is producing energy.

- It fix the fluid amount sync in the generators, some of them were not sending to the client **any amount bigger than 2000 mb**, and the others weren't syncing **any fluid changes**.

- It should reduce the resource comsumption of the ThermalGenerator by checking only once for lava on is sides.

It should be noted that the diesel generator animation is, well, **very special**. 
[Fast animation](http://i.ferenyr.info/images/2016/12/8ce17ac.mp4)
Unaware of the purpose I haven't touched it.

There is also **a major issue** with these generators, they absorb **any fluids** without consulting the recipes. I can fix it in this PR as well if you want.

The semifluid generator **does not have his** activation textures inside the assets, so it become a missing tex.

### DIFF
--- a/src/main/java/techreborn/tiles/generator/TileGasTurbine.java
+++ b/src/main/java/techreborn/tiles/generator/TileGasTurbine.java
@@ -106,31 +106,39 @@ public class TileGasTurbine extends TilePowerAcceptor implements IWrenchable, II
 	@Override
 	public void updateEntity() {
 		super.updateEntity();
-		if (!world.isRemote) {
-			FluidUtils.drainContainers(tank, inventory, 0, 1);
-			tank.compareAndUpdate();
+
+		if (!this.world.isRemote) {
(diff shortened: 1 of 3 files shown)


modmuss50:
Ok, thanks for this. That is an issue if they use any fluids, in 1.7 it used fluid power values from BC and forestry but we dont have them on 1.11. So I think I will have to add my own.

Ourten:
The main problem is that generators will not check if the fluid is valid at all for them. If you place water inside a thermal generator it will produce power just fine. But if water is inserted in a GasTurbine the map is bricked for good.

modmuss50:
Ah right, ok. I will open an issue for this as it seems serious

PR #844, opened 2016-12-15
More Energy Generators fixes

Ourten:
In this PR the changes are **heavily fragmented** in **individual commits**.

- It fix a **continuous activation** with ThermalGenerator if a lava block was adjacent.

- It add a **convenience method** tryAddEnergy(amount) to generators, this method will prevent activation if the Generator is full and fill it to it's real maximum.
The old method was doing that CURRENT + ADD > MAX ? NOT_ADDING : ADDING(ADD) and would make theses weird max values : 393/400.

- It fix the HeatGenerator and DragonEggSiphoner **activation state change**.

- I've also **cleaned up** the HeatGenerator adjacents blocks detection.


modmuss50:
Awesome thanks :+1:

Issue #851, opened 2016-12-17
A lot of issues and questions


davqvist:
Forge 12.18.2.2166
RebornCore 2.8.4.77
TechReborn 1.4.15.127

Hello, I'm working currently on a modpack that plans to use TechReborn as an endgame tech mod.
I don't know a lot about it, so I investigated block after block and had a lot of issues. (Sorry that all the issues are in one bug report, but I don't really feel like making >10 of them (also maybe some aren't even bugs and misunderstandings instead and some might be duplicate to open issues sorry)).

1.) Digital Chest / Quantum Chest: doesn't update numbers and you can't insert items through pipes/hoppers
2.) Sawmill: I was figuring out how it works, it seemed to need liquids, but as I inserted water with an EnderIO Fluid Tank the game crashed and even corrupted my world. I had to delete the block with MCEdit, so this is a major bug. http://pastebin.com/LvsPRqYD
3.) Industrial Grinder: The GUI doesn't update when fluids get added also the recipe in JEI doesn't show if the fluid is water or Mercury.
4.) Transformer: Do they work? I added a MV Transformer behind a MFE to transform down from 512 to 128, machines after it don't get power though (and The One Probe shows no power in the Transformer) and if I place it vice versa it blows up. Also I connected machines that say 128 EU/t max with a MFE with Copper Cables to them and nothing blew up. Is this supposed to happen? Has the MFE the wrong tooltip?
5.) Implosion Compressor / Vacuum Freezer: works but still says "Incomplete Multiblock" even though it is built correctly
6.) Lithium Cell: is there a way to get those? I found a wiki that says that you have to electrolyze Clay Dust to get them, but that doesn't seem to be implemented, just as no way to get Clay Dust
7.) Semifluid Generator / Diesel Generator / Gas Turbine: they don't work with cells, only with buckets but you can't turn cells into buckets as far as I can see. The Diesel Generator works only with Nitrofuel which doesn't seem obtainable, the Semifluid Generator I could only get to work with Biomass, with Creosote the game crashed. http://pastebin.com/j6ZmuGFx All in all those generators seem useless at the moment.
8.) Distillation Tower / Plasma Generator: has no GUI and doesn't connect to cables, so it seems unfinished. Will they be implemented?
[...]

PR #869, opened 2016-12-22
Fix display slots in Fusion Reactor JEI

ghost:
Mentioned in #851, however that issue shouldn't be closed because it mentions many other things as well.


modmuss50:
Haha, not sure how that happened. Thanks :)

PR #951, opened 2017-02-01
Various fixes to cells for 1.7.10

wormzjl:
Tested to be working

### DIFF
--- a/src/main/java/techreborn/init/ModItems.java
+++ b/src/main/java/techreborn/init/ModItems.java
@@ -15,6 +15,7 @@ import techreborn.items.armor.ItemLapotronPack;
 import techreborn.items.armor.ItemLithiumBatpack;
 import techreborn.items.tools.*;
 import techreborn.powerSystem.PoweredItem;
+import ic2.api.item.IC2Items;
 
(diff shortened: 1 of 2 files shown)


modmuss50:
1.7.10 ?!

modmuss50:
I guess I can do one last build, if the jenkins still works...

wormzjl:
Cuz I'm staying with Reika :3

modmuss50:
Are there any other changes for 1.7 that I could add?

wormzjl:
I'd suggest removing the ic2 api from the jar, as it could cause some issues

modmuss50:
Ok I will look into it, it would be nice to leave it in a state where it doesn't break in the future.

wormzjl:
There are also a bunch of recipes I want to fix with regard to gt2, but this could take some time

modmuss50:
Ok, I cannot easily remove the IC2 api from the jar as some of the interfaces are required..

modmuss50:
Ok, I updated all deps and pushed to curse. Also includes your PR from august.

wormzjl:
thx

Issue #992, opened 2017-03-01
Red Garnet dupe from Ore mining


leagris:
### Version:

* TechReborn-1.11.2-2.1.6.76-universal.jar
* Full mod list [here](https://github.com/Beyond-Reality/BeyondRealityModPack/blob/master/BR-1.11.2/modlist.md)

### Detailed description:

When mining a Ruby Ore, a Red Garnet Gem item is looted;
even when Ruby Ore from other mod is mining block instead of breaking it down.
Repeating the process of mining Ruby Ore Block and placing it again leads to Red Garnet Gems duplication exploit.
It may also be possible to exploit extra Red Garnet Gems by silk-touching the ore block from TechReborn.

### Blamed code fragment [in BlockBreakHandler class](https://github.com/TechReborn/TechReborn/blob/1.11.2/src/main/java/techreborn/events/BlockBreakHandler.java#L35-L43):

```Java
if (OreDictUtils.isOre(event.getState(), "oreRuby")) {
	event.getDrops().add(ItemGems.getGemByName("red_garnet").copy());
}
```

### Related issues:
- https://github.com/chazwarp923/Unified-Items/issues/10
- https://github.com/Beyond-Reality/BeyondRealityModPack/issues/446#issuecomment-283365368

modmuss50:
this has caused so many issues. @ProfessorProspector any ideas how to fix it, other wise im removing it.

Prospector:
Just needs a silk touch check. The ruby gem is a normal drop. Must be something else causing that issue

leagris:
Ruby Ore from Unified Items drop the Ore block even if not silk-touched, so there is no reason to loot garnet gems while mining a block without breaking it down.
IMHO you should not try to alter the mining loot of ores from other mods.

Prospector:
We can add a silk touch check but if that doesn't fix the problem, it'll be up to Unified Items to do that. Ores are not meant to drop unless broken with silk touch.

leagris:
I really think it is up to the mod providing a block to take care of what happen when it is broken.
The issue here is that TechReborn intercept the block break event that match any oreRuby Ore Dictionary entry, regardless if the block belong to TechReborn or some other mod; then it spawns a red garnet gem.
- TechReborn has no legitimate role defining the drops from other mod blocks.
- Adding extra loot while intercepting the break event is not the right place to implement loots on breaking an ore block from TechReborn. The red garnet gem should be implemented inside the ore block class from TechReborn.

[...]

PR #993, opened 2017-03-02
Proposed fix for #992

leagris:
1. Check it is not SilkTouching
2. Check the block drop at least a Ruby Gem before adding a Red Garnet to the drops


Prospector:
Oh, yes. Good fix.

PR #1052, opened 2017-04-14
Fix matter fabricator uses scrapboxs when full

jusvit:


modmuss50:
Im not sure why its including your past PR's in the changes, its not a major issue as I can just squish them but you should look into it.

Prospector:
He needs to pull from our branch, because we merged them

jusvit:
@ProfessorProspector How does one do that? I've been trying to get rid of my previous commits, but fell short.

Prospector:
Hm I don't remember, its been a while since I've done stuff on a fork.

PR #1061, opened 2017-04-15
Add tooltips for every* unfinished block/tool

jusvit:
I added the "WIP Coming Soon" for a vast majority of the unfinished blocks. This should prevent any questions arising from the users when they try and use unfinished blocks.

### DIFF
--- a/src/main/java/techreborn/blocks/BlockNuke.java
+++ b/src/main/java/techreborn/blocks/BlockNuke.java
@@ -34,19 +34,15 @@ import net.minecraft.entity.Entity;
 import net.minecraft.entity.EntityLivingBase;
 import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.entity.projectile.EntityArrow;
-import net.minecraft.item.ItemStack;
 import net.minecraft.util.EnumFacing;
 import net.minecraft.util.EnumHand;
 import net.minecraft.util.math.BlockPos;
-import net.minecraft.util.text.TextFormatting;
 import net.minecraft.world.Explosion;
 import net.minecraft.world.World;
 import reborncore.RebornCore;
 import reborncore.common.BaseBlock;
 import techreborn.client.TechRebornCreativeTabMisc;
 import techreborn.entitys.EntityNukePrimed;
-
-import java.util.List;
 
 /**
  * Created by Mark on 13/03/2016.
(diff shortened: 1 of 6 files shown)


src/main/java/techreborn/blocks/BlockNuke.java
```diff
@@ -143,4 +147,10 @@ protected BlockStateContainer createBlockState() {
 		return new BlockStateContainer(this, OVERLAY);
 	}
 
+	@Override
+	public void addInformation(ItemStack stack, EntityPlayer playerIn, List<String> tooltip, boolean advanced) {
```
modmuss50: This works just fine. It has no recipe as it has no sane need in survival.

src/main/java/techreborn/blocks/BlockPlayerDetector.java
```diff
@@ -178,4 +178,11 @@ public IBlockState getStateFromMeta(int meta) {
 	public int getMetaFromState(IBlockState state) {
 		return typeNamesList.indexOf(state.getValue(TYPE));
 	}
+
+	@Override
+	public void addInformation(ItemStack stack, EntityPlayer playerIn, List<String> tooltip, boolean advanced) {
```
modmuss50: Same again, works but needs a recipe.

src/main/java/techreborn/blocks/machine/BlockAssemblingMachine.java
```diff
@@ -84,4 +88,11 @@ public String getTop() {
 	public String getBottom() {
 		return this.prefix + "assembling_machine_top";
 	}
+
+	@Override
+	public void addInformation(ItemStack stack, EntityPlayer playerIn, List<String> tooltip, boolean advanced) {
```
modmuss50: This is not supposed to have a recipe or any recipes

modmuss50 (changes requested):
Either need recipes or the WIP tag removing off the ones I've mentioned

src/main/java/techreborn/blocks/BlockPlayerDetector.java
jusvit: The detect player (All) works but the other two modes do not they act the same as the (All) block

src/main/java/techreborn/blocks/machine/BlockAssemblingMachine.java
jusvit: But it's still a WIP though?

modmuss50: No, it's desgined for pack devs to use.

src/main/java/techreborn/blocks/BlockPlayerDetector.java
modmuss50: That's a bug then, and should be fixed.

jusvit:
Did all the requested changed, ready for merge.

modmuss50 (approved):

Issue #1016, opened 2017-03-17
Fuel is put into the wrong slot when using hoppers with the iron furnace


ghost:
Coal and other fuel items go into the output slot or the top slot (the slot for items that should be smelted/cooked), rather than the bottom slot (the fuel slot) when put in a hopper that is connected to the iron furnace, as opposed to the vanilla furnace, which puts fuel in the right slots.

I am using the latest TechReborn version for 1.11.2.

codewarrior0:
Furthermore, a hopper beneath an Iron Furnace will extract the furnace's fuel instead of its outputs.

TechReborn-1.11.2-2.1.10.88-universal.jar

Issue #968, opened 2017-02-06
[1.11.2] Iron furnace takes fuel into the input slot when shift clicking.


ProsperCraft:
Behavior- shift clicking coal into an iron furnace puts it into the input(to be smelted slot) instead of the fuel slot.

I would expect this in a pack with extreme reactors where there is a recipe for coal to smelt into graphite, but not otherwise.

Expected behavior- shift clicking fuel into the iron furnace puts it into the fuel slot.

PR #1062, opened 2017-04-15
Proposed fix for #1016

jusvit:
Proposed fix for #1016 

Ignore the  #968 fix, will work on that later.

### DIFF
--- a/src/main/java/techreborn/tiles/TileIronFurnace.java
+++ b/src/main/java/techreborn/tiles/TileIronFurnace.java
@@ -77,7 +77,7 @@ public class TileIronFurnace extends TileLegacyMachineBase
 	}
 
 	@Override
-	public void update() {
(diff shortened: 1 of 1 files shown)


modmuss50:
I have a feeling this is an issue with this: https://github.com/Dimmerworld/TechReborn/blob/f94fbb8d08aa08e8e77c8e284392d3b31a5f5457/src/main/java/techreborn/tiles/TileIronFurnace.java#L177

jusvit:
Tried messing with that, to no avail(someone else might have better luck). The current switcharoo seems to work flawlessly(excluding the preexisting quarks with the furnace)

modmuss50:
Humm, ok. I will take a look at the PR another day

modmuss50:
Doing this will also break past tiles. I need to look at the issue in more detail.

PR #1127, opened 2017-06-09
1.7.10 Black GUI Fix

CrossVas:
Hey, it works!!

Explain:
The fix for this is similar with [this](https://github.com/TechReborn/TechReborn/commit/28dd5044e6c9874c0147b48341d10e48291a9ef0)[1.9+]
All you need to do is replacing

**_import net.minecraft.client.renderer.GlStateManager;_**
with 
**_import org.lwjgl.opengl.GL11;_**

and 

**_GlStateManager.color(1.0F, 1.0F, 1.0F, 1.0F);_**
with 
**_GL11.glColor4f(1.0F, 1.0F, 1.0F, 1.0F);_**

in all of machine' GUI'file.


Prospector:
Is it really that bad to use? It's not game breaking, just visual. Plus this has all kinds of merge conflicts. I don't think we really need to do another 1.7 release. The time for that was over a year ago.

modmuss50:
We can do it, it's not harming anyone. I know we shouldn't but I don't see why not. Fix the PR and it should be ok to go

PR #1128, opened 2017-06-10
add dustAsh oreDict recipe

bogger33:
Embers uses dustAsh instead of dustAshes for their Ash oreDictionary. This will make TechReborn ash work with Embers.


modmuss50:
Thanks 👍

bogger33:
slight correction: This doesn't change anything about the PR, but my notes were wrong. It's actually forestry that uses dustAsh, not embers.

PR #1135, opened 2017-06-14
Bug fixes of the fusion reactor.

yuyuyzl:
- Removed the incorrect reset after reaction.
 - Reactor consume input items BEFORE the reaction now.
 - Empty cells will be returned (with a stupid detection)
 - Automation cannot insert to the output slot now.


modmuss50:
Look good, thanks.

PR #1139, opened 2017-06-16
(Another) Bug fixes of the fusion reactor.

yuyuyzl:
- Fixed the energy input settings to fit energy consuming recipes
 - Reactor will reset if the energy is not enough
 - Removed spamming logs.


modmuss50:
Awesome, thanks again.

PR #1232, opened 2017-08-10
Fix hwyla dependency

drcrazy:
Fix hwyla dependency


modmuss50 (changes requested):
Why was the license file removed from ModRecipes??

Looks good once that is fixed.

PR #1247, opened 2017-08-25
Replace all emptiness checks `== ItemStack.Empty` with `.isEmpty()`

### LINKED ISSUE #1157: [1.12] Shift-clicking is weird on TR GUIs
When shift-clicking items to put them in a slot, they transfer but they are still displayed in the player's inventory and they are not displayed in the machine's slot. On the contrary, when shift-clicking to take off items from a TR inventory, they are still displayed in the inventory but not in the player's inventory. The displaying updates when going out of the GUI.

codewarrior0:
Fixes #1157 and probably a few other things.


modmuss50:
Thanks for that, been meaning to do it for a while but didn't get around to it.

codewarrior0:
No problem. Took a bit of debugging to figure out that it was the empty check in `BuiltContainer` that was causing that issue. Figured changing all of them might clear up some other issues, too.

PR #1265, opened 2017-09-06
fix background not showing on somes machines

alexpic57:
add the black background and fix wrong slot position on the chunkloader, iron furnace and scrapboxinator


modmuss50 (approved):
Looks great! Thanks

PR #1290, opened 2017-10-07
Backport commit 08c2f89 to 1.10.2

mracsys:
This allows items to be inserted into/extracted from the fusion reactor in 1.10.2. Copy/paste of the fix for 1.12.

It looks like the affected file was refactored in 1.12. I did not backport those changes.

I had no problems in a test world while letting the reactor run until it's item and energy outputs were full.


modmuss50:
Awesome, thanks, I will get this pushed to curse.

Issue #1306, opened 2017-10-13
Powered lighting


ali1234:
I have implemented a proof-of-concept powered light source for TR:

https://github.com/ali1234/TechReborn/tree/feature/lighting

![lamps](https://user-images.githubusercontent.com/651970/31566373-2cdfe6b6-b062-11e7-91bd-0f70fe128b4d.png)

Currently there are no recipes and power use is not balanced. It is only supposed to accept power through the back side, but currently storage blocks will power it through any side. Cables will connect to any side but only power it from the back.

Unlike the IC2 version, it cannot be turned on and off. It will light as long as it is connected to a power source, so hopefully there will be a switch for cables eventually. Also unlike the IC2 version, this one should be compatible with optifine shader remapping.

I have some ideas for making variants to differentiate it further:

Incandescent lamp:
 * Same brightness as standard torch.
 * Cheap and easy to craft.
 * Relatively high power use. Perhaps it also needs high-voltage T2 power source?
 * Bulky like a traditional light bulb - player can't walk under it if the ceiling is 2 blocks high.

Small LED:
 * Same brightness as glowstone.
 * Significantly less power use than incandescent.
 * A small flat panel with a thin bounding box so the player can walk under it.

Large LED:
 * Uses virtual lights to light a large area.
 * Significantly increased power use.
 * Still has the thin bounding box. Would look like the current implementation.

Grow lights:
 * Same as small LED, but boosts plant growth.

If you don't want to put this in TR I will try to make it a standalone mod. Either way I would still appreciate any feedback on the code and balancing.

modmuss50:
Wow, wasn't expecting a feature PR! Thanks for taking your time.

I really like the idea, and would be happy to see it in the mod. The code looks good and should be ready to merge. 

If you wanted to add the other lamps, I would be happy for you to make a PR adding them as well.

👍

ali1234:
I'm not quite ready to do a PR yet. I just wanted to gauge interest before I polish it up further. :)

drcrazy:
Thank you for your PR. It was really great!

PR #1334, opened 2017-11-04
Implement powered lighting.

ali1234:


ali1234:
This implements #1306.

I'm mostly happy with how it works. Unfortunately I didn't get around to implementing all the different kinds of lights that I wanted, and work has picked up so I won't have time to do any more for a while.

Only two type of light are done:

Incandescent: same brightness as torch. 4 power per tick. (I think 16 EU?). Cheap - uses carbon fibres.

LED: same brightness as glowstone. 1 power per tick. Expensive, uses glowstone dust.

Really the incandescent should use tungsten wire, but tungsten is too expensive and there is no wire form of it.

modmuss50:
Looks great. 👍 Thanks a lot for this.

PR #1391, opened 2017-12-29
Update zh_cn.lang

DYColdWind:


src/main/resources/assets/techreborn/lang/zh_cn.lang
```diff
@@ -375,6 +375,7 @@ item.techreborn.dustsmall.diorite.name=小堆闪长岩粉
 item.techreborn.dustsmall.granite.name=小堆花岗岩粉
 
 item.techreborn.cell.name=空单元
+item.techreborn.cell.fluid.name=$流体$单元
```
modmuss50: This wont work, you will want to have the following:

`item.techreborn.cell.fluid.name=$fluid$单元`

modmuss50 (changes requested):
Almost correct, but this wont work as it only replaces `$fluid$` with the fluid name.

src/main/resources/assets/techreborn/lang/zh_cn.lang
DYColdWind: Thank you for pointing out the mistake ...

liach: Looks funky! modmuss, you should use upper case characters to better indicate such tokens.

modmuss50 (approved):

PR #1408, opened 2018-01-10
Updated to the newer version coming soon (pt-BR)

InterPlay02:
Hello!

Here it is!
But I was thinking that you already have a way to translate that
"Configure Slots" tooltip. Anyway...

See ya!


InterPlay02:
PS.: Added pt-BR.lang file to Reborn Core also! Check it!

PR #1411, opened 2018-01-19
Work on Advanced Drill

temp1011:
Found this and thought I'd give it a go. There are some other bugs with the drill I don't know how to fix though:
- Advanced drill doesn't break blocks in AOE in Creative mode
- Advanced Drill will break adjacent wood/grass/anything else that just requires a hand blocks due to AOE - not necessarily a bug but definitely weird to break wood instantly. I think just need to check if the AOE blocks can be broken with bare hands.
- Might be worth adding blocks to be broken to a list then breaking them all at once so the itemstack produced is together but it doesn't matter too much.

### DIFF
--- a/src/main/java/techreborn/items/tools/ItemAdvancedDrill.java
+++ b/src/main/java/techreborn/items/tools/ItemAdvancedDrill.java
@@ -24,6 +24,7 @@
 
 package techreborn.items.tools;
 
+import codechicken.lib.raytracer.RayTracer;
 import net.minecraft.block.Block;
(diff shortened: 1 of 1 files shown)


drcrazy:
Thank you!

drcrazy:
p1 - due to block breaking with attack and not use and instant block break on attack in Creative I think there is no easy way to overhaul this behavior.
p2 - yeah, weird. I suspect check should be for proper tool, not for bare hands. You can break dirt and gravel and sand with bare hands. And it is fine to break dirt with drill

PR #1413, opened 2018-01-20
More drill work

temp1011:
This is still WIP. At the moment this fixes a bug where using the drill on things like tall grass would still trigger AOE. It seems from what I have tested that `canHarvest()` is broken but I'm not 100%. Still need to stop the drill breaking other stuff when mining dirt etc and get it to work on obsidian etc. I might rewrite more but I'm not really sure at the moment.

### DIFF
--- a/src/main/java/techreborn/items/tools/ItemAdvancedDrill.java
+++ b/src/main/java/techreborn/items/tools/ItemAdvancedDrill.java
@@ -24,6 +24,7 @@
 
 package techreborn.items.tools;
 
+import codechicken.lib.raytracer.RayTracer;
 import net.minecraft.block.Block;
 import net.minecraft.block.state.IBlockState;
 import net.minecraft.creativetab.CreativeTabs;
@@ -35,6 +36,7 @@ import net.minecraft.item.ItemStack;
 import net.minecraft.util.EnumFacing;
 import net.minecraft.util.NonNullList;
 import net.minecraft.util.math.BlockPos;
+import net.minecraft.util.math.RayTraceResult;
 import net.minecraft.world.World;
 import net.minecraftforge.common.MinecraftForge;
 import net.minecraftforge.event.world.BlockEvent;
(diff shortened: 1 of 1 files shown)


src/main/java/techreborn/items/tools/ItemAdvancedDrill.java
```diff
@@ -24,6 +24,7 @@
 
 package techreborn.items.tools;
 
+import codechicken.lib.raytracer.RayTracer;
```
modmuss50, line 27: You cannot use classes from other mods.

src/main/java/techreborn/items/tools/ItemAdvancedDrill.java
```diff
@@ -76,30 +77,42 @@ public boolean canHarvestBlock(IBlockState blockIn) {
...
 
+	public boolean isGoodAOEBlock(IBlockState state) {
+		return !state.getMaterial().isToolNotRequired() || state.getBlock().isToolEffective("pickaxe",state) || state.getBlock().isToolEffective("shovel",state);
+	}
+
 	@Override
 	public boolean onBlockDestroyed(ItemStack stack, World worldIn, IBlockState blockIn, BlockPos pos, EntityLivingBase entityLiving) {
 		if(!(entityLiving instanceof EntityPlayer))
 			return false;
-		RayTraceResult raytrace= rayTrace(worldIn,  (EntityPlayer) entityLiving, false);
```
modmuss50, line 83: I assume my fix didnt work (I didnt test it as I only had my laptop)

modmuss50 (changes requested):
Dont use classes from other mods.

src/main/java/techreborn/items/tools/ItemAdvancedDrill.java
covers1624: Feel free to copy a stripped down version of `RayTracer` from CCL as done in CoFH Core.

temp1011: Thanks, I think I will do this.

modmuss50: What was wrong with my fix? I just tested it and it seemed to work ok?

temp1011: Nothing, this is a bad branch I am commiting from. Let me reset. I got confused with the workspace set up.

temp1011:
Bad branch.

Issue #1415, opened 2018-01-20
Drill Weird Behaviour


temp1011:
I am creating this issue before I start breaking things.

At the moment the Advanced Drill will mine blocks very fast that it shouldn't really.
![before](https://user-images.githubusercontent.com/34900092/35186298-ce7704c4-fe09-11e7-9a10-4921d2537b12.png)
![after](https://user-images.githubusercontent.com/34900092/35186301-d6769b30-fe09-11e7-8623-61f746c00b16.png) With drill aiming at dirt next to logs.
I am aiming to fix this by considering which blocks should be broken.

modmuss50:
I would fix it by making the break speed slower if it contains blocks that shouldn't be broken:

This can be done in the Block Break Event, see an example here: https://github.com/TechReborn/TechReborn/blob/1.12/src/main/java/techreborn/events/BlockBreakHandler.java#L94

temp1011:
Should this carry over to everything eg - drill mining a mix of stone and obsidian will mine at the speed of mining obsidian?

modmuss50:
Yeah, I would say so. But I would ensure that its not going to take ages if you are mining a 9x9 of obsidian as I dont think it needs to be very slow.

temp1011:
I was just planning to break at the speed of the slowest block but I guess the speed drop could be cumulative.

modmuss50:
that would work. Whatever feels right is fine.

temp1011:
I've got to this point but unfortunately I don't get the right effect yet. Any idea what I've done wrong?
 I can PR this but it's not good at the moment. https://github.com/TechReborn/TechReborn/compare/1.12...temp1011:drill?expand=1

modmuss50:
I dont see anything wrong with it just by looking at it. When I get a chance (Later tonight, or later in the week) ill clone it down and take a better look.

My guess is `super.getDestroySpeed(stack,state);` isnt returning the correct speed.

modmuss50:
Any progress on this? or shall I take a look?

temp1011:
I'll have a look again from a fresh perspective. Not sure if I'll get anywhere though.

modmuss50:
Ok, cool. Let me know if you want me to take a peek at this issue. Thanks.

temp1011:
Hmm, still not sure. Something is wrong somewhere but I'm not sure what. I don't know if it's easier/more useful to open a PR or not.

modmuss50:
[...]

PR #1428, opened 2018-01-31
Fix Drill break speed in certain conditions

temp1011:
As discussed in https://github.com/TechReborn/TechReborn/issues/1415. Not sure where the problem is but from what I can tell it doesn't work at the moment.


modmuss50:
Ok, About this, im going to fix it up, and then merge it. You have got the right idea, but with the break speed is a bit off. It is a bit of a pain to make it work as expected.

temp1011:
Great, thanks! Sorry I couldn't get it done right.

modmuss50:
Yeah, ive spent the last hour messing with it trying to get it to feel right, and its still not perfect.

PR #1435, opened 2018-02-10
Update forge and Sound options for Alarm

jusvit:
I've tested it and it doesn't remove existing alarms in the game, therefore should be a fine addition.


src/main/java/techreborn/tiles/TileAlarm.java
```diff
@@ -24,25 +24,68 @@
...
 
 public class TileAlarm extends TileEntity implements ITickable {
-	boolean state = false;
-	int selectedSound;
+	private int selectedSound = 1;
 
 	@Override
 	public void update() {
 		if (!world.isRemote && world.getTotalWorldTime() % 25 == 0 && world.isBlockPowered(getPos())) {
+			System.out.println(selectedSound);
```
modmuss50: This needs to go.

modmuss50 (changes requested):
One small debug line needs to go, and then this looks good.

src/main/java/techreborn/tiles/TileAlarm.java
jusvit: AHHHH

those pesky debugging

PR #1537, opened 2018-06-08
Update ru_ru.lang

yaroslav4167:
Translate rubber plank stairs and slab


modmuss50:
Looks good, thanks for being so quick about it 👍

PR #1552, opened 2018-06-19
Advanced Jackhammer Recipe

electromatter:
During the "Great Refactoring" the crafting recipe for the advanced jackhammer was not correctly entered. The current version has a diamond drill in place of a diamond jackhammer to upgrade to the advanced. See commit 6d931b232d0670a8ccabab5fc596177312b5a8c0 for the original recipe on line src/main/java/techreborn/init/ModRecipes.java:966. Which ended up at src/main/java/techreborn/init/recipes/CraftingTableRecipes.java:80 after the refactoring. The name of the advanced jackhammer also sticks out as it is "ironjackhammer" as opposed to the expected "advancedjackhammer"

I recommend two changes:

1.  Correcting the recipe
2.  Correcting the name of the advanced jackhammer to be "advancedjackhammer"

Renaming "ironjackhammer" => "advancedjackhammer" is a potentially breaking change if any mod packs or texture packs refer to the item by name. However I judge this to be unlikely. I have grep'ed this repository and find no other references to "ironjackhammer".

### DIFF
--- a/src/main/java/techreborn/init/ModItems.java
+++ b/src/main/java/techreborn/init/ModItems.java
@@ -224,7 +224,7 @@ public class ModItems {
 		DIAMOND_JACKHAMMER = new ItemDiamondJackhammer();
 		registerItem(DIAMOND_JACKHAMMER, "diamondjackhammer");
 		ADVANCED_JACKHAMMER = new ItemAdvancedJackhammer();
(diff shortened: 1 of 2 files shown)


drcrazy:
rename cannot happen untill 1.13
the recipe fix is fine

Please, undo name change.  

I've added your PR to list of things to fix in 1.13 so name will be changed there.

PR #1553, opened 2018-06-19
Corrected the Advanced Jackhammer recipe

electromatter:
This is the change referred to by #1552 with only the recipe change and not the rename.


drcrazy:
Merged. Thank you for your contribution!

PR #1556, opened 2018-06-23
Fusion reactor consumes resources. Fixes #1551

### LINKED ISSUE #1551: Fusion Control Computer not using up input
Okay so as the title suggest the Fusion Control Computer is not using up its fuel - atleast in my case.

I am currently playing on FTB Continuum 1.3.1 (which is currently the latest version).
The version of TechReborn is

_TechReborn-1.12.2-2.15.15.712-universal.jar_

So it happened like this:
I added two hoppers on both sides (so 4 in total) and put Deuterium and Helium3 in them. I then configured the Slots of the Fusion Control Computer. After I saw that it was putting it in I just added that the resulting Heliumplasma would get extracted in a chest and thats it. I went to sleep. The next day the Reactor was still running and did not use up any of the Deuterium and Helium3 cell but ran all the time. It created way more Heliumplasma cells than I even had Helum3 and Deuterium combined. Now looking at it, it is still not using any of the cells.

I was also able to reproduce it by just putting a stack of each into the Computer, it used one, and now it is just running on "air"

http://prntscr.com/jwbky7

This issue is similar to issue #1539 
but as you can see just happens for no apparent reason

If you need any more help or you are not able to reproduce the error don't hesitate to ask me.

electromatter:
Manually tested. Corrected by consuming the resources when the machine restarts the crafting cycle.

### DIFF
--- a/src/main/java/techreborn/tiles/fusionReactor/TileFusionControlComputer.java
+++ b/src/main/java/techreborn/tiles/fusionReactor/TileFusionControlComputer.java
@@ -259,6 +259,10 @@ public class TileFusionControlComputer extends TilePowerAcceptor
 					}
 					if (this.validateReactorRecipe(this.currentRecipe)) {
 						this.crafingTickTime = 0;
+						this.decrStackSize(this.topStackSlot, this.currentRecipe.getTopInput().getCount());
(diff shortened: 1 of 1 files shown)


drcrazy:
Thank you!

PR #1557, opened 2018-06-23
Some compability crash preventing

VimbaTheFish:
Added check for casting item from slot to IEnergyItemInfo (prevents crashes)

### DIFF
--- a/src/main/java/techreborn/tiles/storage/TileEnergyStorage.java
+++ b/src/main/java/techreborn/tiles/storage/TileEnergyStorage.java
@@ -72,13 +72,15 @@ public class TileEnergyStorage extends TilePowerAcceptor
 		super.update();
 		if (!inventory.getStackInSlot(0).isEmpty()) {
 			ItemStack stack = inventory.getStackInSlot(0);
-			if (!(stack.getItem() instanceof IEnergyItemInfo)){
-				return;
-			}
 			if (stack.hasCapability(CapabilityEnergy.ENERGY, null)) {
 				IEnergyStorage powerItem = stack.getCapability(CapabilityEnergy.ENERGY, null);
-				int maxReceive = Math.min((powerItem.getMaxEnergyStored() - powerItem.getEnergyStored()),
-						(int) ((IEnergyItemInfo) stack.getItem()).getMaxTransfer(stack));
+				int maxReceive;
+				try {
+					maxReceive = Math.min((powerItem.getMaxEnergyStored() - powerItem.getEnergyStored()),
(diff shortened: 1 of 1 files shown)


modmuss50:
No, this will break charging none TechReborn items. Have you got an example crash?

modmuss50 (changes requested):
This shouldn’t need to cast anything as it uses the forge api (our items support this). Where is it crashing?

VimbaTheFish:
Yep, my fault. Didn't thought about this.
Tried to fix that: https://pastebin.com/34vaw4HV - Iron Jetpacks mod. Crashes when u try to charge it.
The same charging works in EnderIO.

Anyway, I don't know does even this problem from TechReborn side : (

modmuss50:
I have a feeling they are using a old version of TechReborn as all this code was changed recently. Can you find what version is being used? Thanks.

VimbaTheFish:
That's standalone mod, uses ForgeEnergy. I don't see in build.gradle any TR dependencies.
https://minecraft.curseforge.com/projects/iron-jetpacks
https://github.com/BlakeBr0/IronJetpacks

drcrazy:
Changes in powered items were released in TechReborn-1.12.2-2.15.10.677 on 27-MAY-2018

modmuss50:
Ah I see the cast now. That is a horrible fix, it should be fixed to work with all mods.

SkyFire55:
I made try-catch for ClassCastException for me, coz I need it to prevent crashes (at least). SHould I create new pull request?

VimbaTheFish:
Yup, that's horrible fix. I'll use it untill it will be fixed anyway, have no choise : )

modmuss50:
No, don’t try catch it. Remove the cast to the powered item and use an api that forge provides. Failing that it seems to want the max transfer of the item, I have a feeling forge items handle this so it can be ingnored. Sorry for not being too helpful as I’m only on my phone atm.

drcrazy:
I can add check for instance of IEnergyItemInfo in that particular place. Or rewrite to simulate first. What is preferable way?

modmuss50:
I would change the get max transfer (the only one being cast that I can see) to use the max input of the machine. This then allows for the item to limit it more if it wants.

VimbaTheFish:
Should I do anything? I can replace cast place on machine maxOutput int

VimbaTheFish:
Yup, it works, tested it recently. Thanks for help!

drcrazy:
I'm closing this now. It is fixed by https://github.com/TechReborn/TechReborn/commit/af6cb38bfba3ff740189798042cab58c8a7019b6 in a little bit different way. 

Anyway, thank you a lot for detailed analysis of the crash cause.

PR #1585, opened 2018-08-22
1.13 prep - Plates

drcrazy:
Initial flattening of TR plates


src/main/java/techreborn/init/ModPlates.java
```diff
@@ -0,0 +1,92 @@
...
+import net.minecraft.util.IStringSerializable;
+import techreborn.lib.ModInfo;
+import techreborn.items.ItemPlates;
+
+
+/**
+ * @author drcrazy
+ *
+ */
+public enum ModPlates implements IStringSerializable {
```
modmuss50, line 37: Looks like a good start.

Would it be possible to set the camel case name instead of having to duplicate it twice?

`ADVANCED_ALLOY_PLATE("plateAdvancedAlloy"),`

would then just be

`ADVANCED_ALLOY_PLATE,`

src/main/java/techreborn/init/ModPlates.java
```diff
@@ -0,0 +1,92 @@
...
+	public final Item item;
+	
+	private ModPlates(String name) {
+		this.name = name;
+		this.item = new ItemPlates();
+		// ModItems will take care about setRegistryName
+		//this.item.setRegistryName(new ResourceLocation(ModInfo.MOD_ID, name));
+		this.item.setTranslationKey(ModInfo.MOD_ID + "." + name);
+	}
+
```
modmuss50, line 87: I would add a getter for the item + item stack.

In 1.13 there we can use ItemProvider as well so this will be really good.

src/main/java/techreborn/init/ModItems.java
```diff
@@ -165,8 +164,11 @@ public static void init() {
 		registerItem(DUSTS, "dust");
 		SMALL_DUSTS = new ItemDustsSmall();
 		registerItem(SMALL_DUSTS, "smallDust");
-		PLATES = new ItemPlates();
-		registerItem(PLATES, "plates");
+
+		for (ModPlates plate : ModPlates.values()){
```
modmuss50, line 168: This can be moved into the enum, and then possibly have an enum of enums. So its only one specific call.

src/main/java/techreborn/client/RegisterItemJsons.java
```diff
@@ -131,11 +133,10 @@ private static void registerItems() {
 			registerBlockstate(ModItems.GEMS, i, name[i], "items/materials/");
 		}
 		
-		name = ItemPlates.types.clone();
-		for (int i = 0; i < ItemPlates.types.length; ++i) {
-			registerBlockstate(ModItems.PLATES, i, name[i], "items/materials/");
+		for (ModPlates plate : ModPlates.values()){
```
modmuss50, line 136: Move to a SideOnly(Client) method in the enum, same as registering the items.

modmuss50:
Great start 👍 

A few pointers on what to do next. Its going to be a massive amount of work to move everything to this, but its looking good.

src/main/java/techreborn/init/ModPlates.java
drcrazy: Options I've considered:
1) "PLATE_ADVANCED_ALLOY," this could be easily case converted to oreDict,
2) "ADVANCED_ALLOY,", this could be prefixed and case converted to oreDict 

Moreover, since not a lot left in ItemPlates class maybe we will end up with ItemMaterial, which will accept allowed prefixes either in constructor or via hasPlate("true") method.

src/main/java/techreborn/init/ModPlates.java
drcrazy: item is public property. Any specific reasons to have getter for item? 
ItemStack is a good idea.

src/main/java/techreborn/client/RegisterItemJsons.java
drcrazy: ok

src/main/java/techreborn/init/ModPlates.java
modmuss50: I think just `ADVANCED_ALLOY` would be good, as when referencing it it would be `ModPlates.ADVANCED_ALLOY` no need to duplicate the plate bit. 

If you did extra data for them you could have a plates.json file somewhere that has stuff like the oredict name and what not. (Thinking about it is not a great idea)

But it would be really nice to have everything in oneplace.

PR #1588, opened 2018-08-24
Fix for Rock Cutter getting an enchant on every re-craft

YuRaNnNzZZ:


modmuss50:
Looks nice and simple. Thanks a lot.

PR #1643, opened 2018-12-27
Remove almost all explicit uses of CapabilityEnergy.ENERGY

coderbot16:
As it turns out, most of the uses of `CapabilityEnergy` are Tech Reborn items accessing their own power supply. In these cases it is safe to assume that getCapability will return a ForgePowerItemManager, meaning that instead of going through the capability system items can just directly access their power supply. In some cases this eliminates a cast where the implementation already assumed it was using ForgePowerItemManager.

This makes it possible to use FE internally, while adding the possibility of not exposing FE externally. Eventually this will become useful for enabling a true EU-only mode of TechReborn, while also providing the option to have TechReborn supporting FE. 

For now, this does not add or remove any functionality, but it is simply a refactoring to open up the possibility of disabling the returning CapabilityEnergy.ENERGY from PoweredItemContainerProvider.

However, there is one hole with the changes: the Lithium Backpack and Lapotron Pack still have to use CapabilityEnergy directly since the only way to charge items through the IC2 Helper is from a TilePowerAcceptor. Therefore, I think that it would work to add another API to charge an item from a ForgePowerItemManager combo. Also, StackInfoHud and ItemUtils need to be modified in RC to use ForgePowerItemManager directly as well before CapabilityEnergy returning can be disabled.

### DIFF
--- a/src/main/java/techreborn/blocks/BlockRubberLog.java
+++ b/src/main/java/techreborn/blocks/BlockRubberLog.java
@@ -41,10 +41,10 @@ import net.minecraft.util.SoundCategory;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.world.IBlockAccess;
 import net.minecraft.world.World;
-import net.minecraftforge.energy.CapabilityEnergy;
 import net.minecraftforge.energy.IEnergyStorage;
 import prospector.shootingstar.ShootingStar;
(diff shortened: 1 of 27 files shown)


modmuss50:
This looks good.

Making those changes should be easy enough

PR #1660, opened 2019-01-29
Transformers now declare that they can output multiple packets per tick

coderbot16:
Should be simple enough.


modmuss50:
Looks good

PR #1687, opened 2019-03-12
Fix RollingMachine tile not updating its active state

ralphII:


modmuss50:
Looks good, thanks for this.

PR #1729, opened 2019-07-12
Implement IMiningDrill for IC2 miner compatibility with TR Drills

estebes:
Also bumping forge version because of JEI


modmuss50 (approved):
Looks great, thanks a lot for this

PR #1730, opened 2019-07-13
Eu support for cables

estebes:
Also better cable connectivity logic.


src/main/java/techreborn/tiles/cable/TileCable.java
```diff
@@ -254,10 +310,22 @@ public void addInfo(List<String> info, boolean isRealTile) {
...
-	// IToolDrop
+	// IToolDrop >>
 	@Override
 	public ItemStack getToolDrop(EntityPlayer playerIn) {
 		return cableType.getStack();
 	}
+	// << IToolDrop
+
+	// Fields >>
+	public int power = 0;
```
modmuss50: Generally we have the fields at the top of the class

src/main/java/techreborn/tiles/cable/TileCableEU.java
```diff
@@ -0,0 +1,321 @@
...
+import reborncore.common.RebornCoreConfig;
+import reborncore.common.powerSystem.PowerSystem;
+import reborncore.common.util.StringUtils;
+import techreborn.blocks.cable.BlockCable;
+import techreborn.blocks.cable.EnumCableType;
+
+import java.util.List;
+
+/**
+ * Created by modmuss50 on 19/05/2017.
```
modmuss50: Put your name here if you want :)

src/main/java/techreborn/blocks/cable/BlockCable.java
```diff
@@ -145,7 +149,11 @@ public EnumBlockRenderType getRenderType(IBlockState state) {
 	@Nullable
 	@Override
 	public TileEntity createNewTileEntity(World worldIn, int meta) {
-		return new TileCable(getStateFromMeta(meta).getValue(TYPE));
+		if (Loader.isModLoaded("ic2") && PowerSystem.EnergySystem.EU.enabled.get()) {
```
modmuss50: Might not be a bad idea to add a config option for this, as it could change the behavour of the cables in existing worlds.

Right now people could be using our cables to transfer RF power when IC2 is installed. I do think its out of scope and not a good idea to have a cable to works with both EU and FE, but having a config to force FE on the cables might not go amiss.

src/main/java/techreborn/tiles/cable/TileCableEU.java
```diff
@@ -0,0 +1,321 @@
...
+	 */
+	@Optional.Method(modid = "ic2")
+	@Override
+	public boolean emitsEnergyTo(IEnergyAcceptor iEnergyAcceptor, EnumFacing enumFacing) {
+		return true;
+	}
+	// << IEnergyConductor
+
+	// Fields >>
+	private int transferRate = 0;
```
modmuss50: and again fields at the top of the class :)

modmuss50 (changes requested):
Overall the code looks good, 👍  I can see it was a real pain to do. Having 2 tile classes does seem the best way to go.

I would add a config (default off) to allow the forcing of FE cable mode as I can see it changing the way the cables work for some people.

This may break old builds, but we can add a warning to it, and I think it will be worth it.

We usually like to have the fields at the top of the classes, that should be an easy change.

modmuss50 (approved):

modmuss50:
Thanks a lot for this 👍

PR #1745, opened 2019-08-07
(1.14 Fabric) Moved things from assets to data

Shnupbups:
Moved advancements and the rubber plantation loot table to data, where they should be


modmuss50:
Ah, nice spot. Thanks for this.

PR #1793, opened 2019-08-18
Update to REI 3.0-pre

shedaniel:
This will update TechReborn to REI 3.0. Also made the label draw without a shadow (It will look better that way)


modmuss50:
Is 3.0 released? 

PR looks good just want to merge this at the correct time so people arent forced onto the old version.

shedaniel:
Yep it is, I am PRing to mods to fix this. (3.0 was out because I need to fix `java.awt`)

shedaniel:
The old entrypoint works too. But by changing to `rei_plugins_v0`, the mod won't crash in REI 2.x.

modmuss50:
Thanks a lot for this.

PR #1851, opened 2019-10-09
Add LibCD Support

LemmaEOF:
Very WIP. Will require at least one PR to RebornCore to remove the necessity of the ugly hacks I use here.

This PR adds LibCD Tweaker support for (almost) all TechReborn recipes. This means that users will be able to use JSR-223 compatible scripting languages to add recipes without writing JSON. This is immensely helpful for modpack devs, since they can programmatically create whole sets of recipes with a couple lines of code instead of having to provide large numbers of JSON files. LibCD is soft depended on using `FabricLoader.getInstance().isModLoaded()` (as LibCD doesn't have an entrypoint yet), so it will not cause crashes if not present.

Still needs finishing:
- [x] custom ingredient serializer class as to not require the hacks that `WrappedIngredient` currently uses
- [x] add Rolling Machine recipes which support [all four forms of LibCD shaped recipes](https://github.com/CottonMC/LibCD/blob/master/src/main/java/io/github/cottonmc/libcd/tweaker/RecipeTweaker.java#L168-L239)
- [x] PR to RebornCore to add ways to construct RebornRecipes other than JSON deserialization as to not require the hacks that `TweakedRebornRecipe` currently uses
- [x] add support for Fluid Generator recipes


LemmaEOF:
Alright, LibCD support is fully implemented! I'm gonna write some docs and sample code soon.

LemmaEOF:
Alright! I have the sample script fully up and running. Everything works just fine as far as I can test (#1852 and a LibCD bug notwithstanding), and every script-facing method has javadocs! I also added in a `TweakerStackGetter` for fluid cells, so a script author can get a filled cell without having to do the NBT themselves!

src/main/java/techreborn/compat/libcd/TRRecipeParser.java
```diff
@@ -0,0 +1,102 @@
...
+                    Identifier id = new Identifier(in);
+                    Item item = Registry.ITEM.get(id);
+                    if (item == Items.AIR) throw new TweakerSyntaxException("Failed to get item for input: " + in);
+                    stack = new ItemStack(item);
+                }
+                return new StackIngredient(Collections.singletonList(stack), amount, tag, requireEmpty);
+            }
+        }
+        else throw new TweakerSyntaxException("Illegal object passed to TechReborn of type " + input.getClass().getName());
+    }
```
modmuss50, line 89: Quite a bit of code here that seems like it should be in libcd. Is there a specific reason why this needs to be in TR? (Thinking of the maintenance burden)

src/main/java/techreborn/utils/ShapedRecipeSerializer.java
```diff
@@ -0,0 +1,71 @@
...
+import com.google.gson.JsonObject;
+import net.minecraft.item.ItemStack;
+import net.minecraft.recipe.Ingredient;
+import net.minecraft.recipe.ShapedRecipe;
+import net.minecraft.util.registry.Registry;
+
+import java.util.HashMap;
+import java.util.Map;
+
+//TODO: switch RebornRecipes over to proper packet-based serialization instead of JSON packets
```
modmuss50: I will try and find some time to do this before merging this PR.

modmuss50 (commented):
Looks good overall. 1 query about the code and I think cleaning up the recipe syncing is a good idea to do before pulling this in.

src/main/java/techreborn/compat/libcd/TRRecipeParser.java
LemmaEOF: This *is* effectively in LibCD already, but TR needs its own copy because it doesn't use vanilla ingredients.

modmuss50: Humm yeah, good point. What about using the wrapped indgredient? If not this isnt too much.

LemmaEOF: The primary issue is that vanilla ingredients only accept one item at a time; you can't set a count like you can in TR. That's a big problem with vanilla in general, but I'm not sure how feasible it would be to have LibCD mix into all the vanilla stuff to fix it.

modmuss50: Ah ok thats a good point. Dont worry about it its not like its 1000s of lines.

src/main/java/techreborn/utils/ShapedRecipeSerializer.java
modmuss50: Should be sloved now, let me know if not.

LemmaEOF:
Alright, everything looks good! This should be ready to go.

LemmaEOF:
I *will* get rid of the ShapedRecipeSerializer quickly, though, since it's no longer needed.

LemmaEOF:
Actually, wait. Not ready just yet; the `RollingMachine.serialize(PacketByteBuf)` and `RollingMachine.deserialize(PacketByteBuf)` need to be put somewhere else so they can be properly used. Just a minute.

LemmaEOF:
Hmmmm. This is gonna take a bit more work than I thought. RebornRecipes themselves don't have a serialize/deserialize with a packetbytebuf, so I can either:

1. add that to RebornRecipes and give it default behavior (probably the best option)

2. add an extension of RebornRecipeType to RollingMachineRecipe

modmuss50:
Im not sure why you need access to those? can you jump on irc as its easier to talk?

LemmaEOF:
Turns out the issue was my local RC copy wasn't up to date. Everything's good now!

modmuss50 (approved):
Im happy with this now, thanks for the help.

I will merge once you give the OK :)

LemmaEOF:
Added one last change to make the sample script log when it's run. It shouldn't be possible for the script to run if LibCD dev mode isn't enabled, but just wanna make sure. It's merge-ready!

modmuss50:
Thanks for this :) Mind if I use the sample script on a wiki page?

LemmaEOF:
Go ahead! That's what it's for. It's got at least one registered recipe for each TR type, along with example calls to creating fluid ingredients and using the cell stack-getter.

PR #1855, opened 2019-10-15
Fix tweaker_test.js.mcmeta

LemmaEOF:
I just found a bug with LibCD where a malformed condition mcmeta would cause the sample script to load in production environments. The bug is fixed with LibCD 1.4.0, but I realized that the TR `tweaker_test.js.mcmeta` is malformed in the way that causes the bug. This fixes the mcmeta so the script will actually run in dev mode.

### DIFF
--- a/src/main/resources/data/techreborn/tweakers/tweaker_test.js.mcmeta
+++ b/src/main/resources/data/techreborn/tweakers/tweaker_test.js.mcmeta
@@ -1,5 +1,5 @@
 {
-    "when": {
-        "libcd:dev_mode": true
-    }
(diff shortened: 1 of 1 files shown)


modmuss50:
Does this require a new release to curse? Or is production un effected?

LemmaEOF:
In production, instead of loading the tweaker script when it shouldn't, it'll now log that there's a malformed `when` block for the mcmeta.

modmuss50:
Ok, I’ll push a release anyway in a bit. It’s only a simple commit.

PR #1901, opened 2019-12-07
Add comparator support for energy storage and generator blocks

Juuxel:
Requires TechReborn/RebornCore#132, which also provides comparator support for machines.

Adds comparator output to energy storages, calculated as `ceil(energy * 15.0 / totalEnergy)`.


modmuss50 (approved):
Looks good, thanks for this.

PR #1902, opened 2019-12-10
update to LibCD 2.0

LemmaEOF:
Updates the LibCD compat to LibCD 2.0. Main changes:
- uses the new LibCD entrypoint for registration, meaning no `FabricLoader.getInstance().isModLoaded()` check in the main entrypoint
- recipe-add methods now take an `Object[]` for outputs instead of an `ItemStack[]`
  - due to this change, the `TRTweaker.addRollingMachine(String[], Map<String, Object>, ItemStack)` method has been renamed to `TRTweaker.addDictRollingMachine(String[], Map<String, Object>, ItemStack)` to prevent Nashorn method overload. This is a breaking change that cannot be put through deprecation.
- the ancient bug causing `TRTweaker.addRollingMachine(Object[], Object, int, int)` to throw `IndexOutOfBoundsException`s has been fixed
- running `/libcd debug export` now adds info about fluid generator recipes 
- `TRTweaker` now uses a LibCD logger instead of a TechReborn one, so that the namespace of the script registering recipes can be displayed
- the tweaker sample script has been updated for all new changes

Known issues:
- due to RebornCore not registering `RebornRecipeType`s in `Registry.RECIPE_TYPE`, any RebornRecipe added will show up as the `unknown` recipe type in RecipeTweaker's debug export. I am not willing to add special behavior for RebornRecipes.
- the LibCD API, while now fully self-contained, does not currently have a JiJ-able form. This will be remedied when I have the time.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -82,10 +82,10 @@ dependencies {
 
 	modCompileOnly "io.github.prospector:modmenu:1.7.9-unstable.19w34a+build.1"
 	modCompileOnly ("me.shedaniel:RoughlyEnoughItems:3.2.7-unstable.201911150440")
-	modCompile ('io.github.cottonmc:LibCD:1.5.0+1.14.4') {
+	modCompile ('io.github.cottonmc:LibCD:2.0.1+1.15') {
 		transitive = false
 	}
-	modRuntime ("io.github.cottonmc:Jankson:1.0.0+j1.1.2") {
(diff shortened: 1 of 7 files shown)


modmuss50:
Looks good, when I get a chance ill take a look at the RebornRecipeTypes not being registered.

modmuss50 (approved):

PR #1909, opened 2019-12-15
Remove mod menu from pom

natanfudge:


modmuss50 (changes requested):
The ideal fix will be to use MM of my maven, after MMs pom has been fixed.

natanfudge:
Either way you don't want to include optional dependencies with `modCompile` (it should be `modRuntime`), same thing with REI.

modmuss50:
This wont work for REI as it compiles against it.

natanfudge:
In that case you need to add `modCompileOnly` additionally. But still not `modCompile`.

modmuss50:
That will then stop it being there when ran in dev... There are better ways to control what goes into the pom.

natanfudge:
> That will then stop it being there when ran in dev...

No, because you will _also_ have the `modRuntime`. For optional dependencies you want to compile against, you use _both_ `modRuntime` _and_ `modCompileOnly`. What better way is there to do it?

modmuss50:
You can use `pom.withXml` in the MavenPublication to directally control what the pom contains. In this case i think thats a better way to do it.

natanfudge:
Can you _remove_ xml entries with that? Either way tinkering with the xml is not a better solution for [literally what compileOnly is made for](https://blog.gradle.org/introducing-compile-only-dependencies).

modmuss50:
Having duplicate entrys isnt ideal.

natanfudge:
In that case you can do 
```
configurations {
    optionalDependency {
        extendsFrom modRuntime
        extendsFrom modCompileOnly
    }
}

dependencies {
     optionalDependency "REI"
}

```

**Update:** It seems like this doesn't work for some reason, perhaps a Loom bug. Invoking them both individually still works. So in Kotlin DSL you can do (not really relevant in TR):
```kotlin
fun DependencyHandlerScope.optionalDependency(dep: String) {
   modRuntime(dep) {
        exclude(group = "net.fabricmc.fabric-api")
   }
   modCompileOnly(dep)
}
```

modmuss50:
Yeah, that might be a better idea actaully.

natanfudge:
See edit^

modmuss50:
Closing the PR as its not the right solution. Ill look into a full fix.

Issue #1913, opened 2019-12-18
Suggestion - Better in-game & wiki documentation


Trnrr:
Just to give a quick perspective of where I'm coming from. Others might not have this issue. I haven't played a survival modded server with friends since around 2015 and recently got back into it with a few friends. We found fabric and decided to make a small pack to play. So I'm not as experienced with some of the tech mods and how they function so please bare with me 😄 

We use the latest fabric version 1.15.1 and latest tech reborn that was updated today.

**Wiki:**
Some stuff off the wiki is outdated or has little info. It would be nice if this was eventually updated. For instance, iron furnace and all the upgrades are missing. It's somewhat hard to try and figure out what some things do because of this and the bad in-game documentation.

**In-game:**
It would be nice if the machines had a basic tooltip of what they did. Same with the upgrades. It would be nice to show their pros/cons because right now I just have to try to make an educated guess. Overclocker prob makes it faster but use more energy? Not sure about transformer or superconductor. Energy storage prob just increases energy storage.

Some solutions could be either a book that explains the machines sort of like the tinkers book. Another could be tooltips that explain what the machines do. Some mods we used to play if you held shift it would show more more for the tooltip.

I know that fabric is still newer and the modding scene is still growing so if some of the stuff isn't possible yet that is ok. I just wanted to give a suggestion from a newer player trying to learn techreborn.

Thank you for making it for 1.15. We were glad when we found a 1.15 tech mod since that enabled us to play on 1.15!

jusvit:
I'll add some tooltips for the upgrades and something potentially for machines, however the machines name kinda imply what they're used for. I'll look into it.

modmuss50:
The wiki is an on going issue, it takes a lot of work to maintain. Im happy for people to help keep it updated (poke me on the discord server)


Tooltips are a great idea, and @justinvvitale has a PR open to slove this.

A book is also a lot of work, as it requires coding and then also requires the documatation to be written.

Trnrr:
I appreciate it! I will message you on discord if I think of anything that could be added to the in-game documentation or wiki. I would offer to help maintain it but I just don't know enough about the mod yet.

modmuss50:
Should be better now with the new tooltips. The wiki is an ongoing effort by all.

PR #1917, opened 2019-12-22
#1913 Improved in-game documentation via tooltips for both machines & upgrades and solar panel exception fix

jusvit:
Added whole bunch of tooltips for common machines, please read through them, mostly derived from the wiki and looking at the block in-game. Also had to keep them shortish so they don't flow off the screen 

Added a default constructor for the solar panel, was throwing exceptions when loading existing world.

Also organised the imports in things I've previously touched

#1913


src/main/java/techreborn/items/ItemUpgrade.java
```diff
@@ -64,15 +54,4 @@ public void process(
 			ItemStack stack) {
 		behavior.process(blockEntity, handler, stack);
 	}
-
-	@Environment(EnvType.CLIENT)
-	@Override
-	public void appendTooltip(ItemStack stack, @Nullable World world, List<Text> tooltip, TooltipContext context) {
-		if(stack.getItem() == TRContent.Upgrades.SUPERCONDUCTOR.item){
-			if(Screen.hasShiftDown()){
-				tooltip.add(new LiteralText(Formatting.GOLD + "Blame obstinate_3 for this"));
```
modmuss50, line 73: Can we keep this please. I know its an "in joke" but its something that I would like to keep in the mod. Thanks.

modmuss50 (changes requested):
Awesome work here, just one thing that I would like kept in. Thanks.

modmuss50 (approved):
Looks good, thanks a lot again 👍

PR #1923, opened 2019-12-24
[Proposal] Solar panels dustiness mechanic

vhd:
Solar panels will require some maintenance to be efficient. This will force player to build some contraptions around them with pistons and redstone.

In action https://youtu.be/wdFnNp1hLUY


modmuss50:
Its a nice idea, but I think it may be a bit too much for the normal player. Id be happy to include something like this if it was disabled by default. (pack devs/players can enable it if wanted)

vhd:
After some playing i have mixed feelings about it.
Right now i'm not sure it worth to be merged.
If this PR doesn't bother you then let it stay on the list

PR #1928, opened 2019-12-27
[Proposal] Greenhouse Controller (aka Farming Station)

vhd:
Demo here https://youtu.be/EsvoXEN_ZjA


modmuss50:
Wow! thats awesome. I was 1/2 thinking of adding a farming block 👍

src/main/java/techreborn/blockentity/machine/tier1/GreenhouseControllerBlockEntity.java
```diff
@@ -0,0 +1,236 @@
...
+				world.breakBlock(blockPos.up(y), false);
+			}
+		}
+		
+	}
+	
+	private void insertIntoInv(List<ItemStack> stacks) {
+		stacks.forEach(stack -> {
+			for (int i = 0; i < 6; i++) {
+				if (insertIntoInv(i, stack).isEmpty()) {
```
modmuss50: Might be worth making it only work if it has free space in the inv. as this will just void the items once full I think?

vhd:
Baby steps into Java, modding and ... english!
So, please, double check all what i'm doing  :)

src/main/java/techreborn/blockentity/machine/tier1/GreenhouseControllerBlockEntity.java
vhd: Yep, that's correct. That was in one of my TODO lists that i forget about :(

src/main/java/techreborn/blockentity/machine/tier1/GreenhouseControllerBlockEntity.java
```diff
@@ -0,0 +1,236 @@
...
+			for (int y = 1; world.getBlockState(blockPos.up(y)).getBlock() == block; y++) {
+				if (canUseEnergy(TechRebornConfig.greenhouseControllerEnergyPerHarvest)) {
+					useEnergy(TechRebornConfig.greenhouseControllerEnergyPerHarvest);
+					insertIntoInv(Block.getDroppedStacks(blockState, (ServerWorld) world, blockPos.up(y), null));
+				}
+				world.breakBlock(blockPos.up(y), false);
+			}
+		}
+		
+	}
```
modmuss50: There is a fair amount of duplicated code here, 

```java
if (canUseEnergy(TechRebornConfig.greenhouseControllerEnergyPerHarvest)) {
	useEnergy(TechRebornConfig.greenhouseControllerEnergyPerHarvest);
	insertIntoInv(Block.getDroppedStacks(blockState, (ServerWorld) world, blockPos, null));
	world.setBlockState(blockPos, blockState.with(SweetBerryBushBlock.AGE, 1), 2);
}
```

is mostly the same across all of them. I would make a basic api out of it to slove the duplicate code and the isuse of the voiding items.

Something like this could work well. (made in 30 seconds, could do with some improvements)

```java
import net.minecraft.block.BlockState;
import net.minecraft.item.ItemStack;
import net.minecraft.util.math.BlockPos;
import net.minecraft.world.World;

import java.util.List;

interface CropHandler {

	/**
	 * Used to break a crop block and return its drops. Returns a list of item stacks, can be empty but never null.
	 * removeFromWorld can be used to check if the inv has space before removing from the world
	 * 
	 * @param world The {@link World}
	 * @param blockPos The {@link BlockPos} of the crop, may not be the block to break (sugar cane 1 up)
	 * @param blockState the {@link BlockState} of the block
	 * @param removeFromWorld when true the block is removed from the world, when false it still returns the items to allow checking for output slots.
	 * @return A {@link List} of {@link ItemStack} to insert into the inventory
	 */
	List<ItemStack> tryBreak(World world, BlockPos blockPos, BlockState blockState, boolean removeFromWorld);
}
```

drcrazy: LogBlock harvest missing. 
Vanilla saplings are instance of Fertilizable. Thus I assume that vanilla trees are supported.

drcrazy:
Офигитительно!

modmuss50:
Trees will need special casing to ensure they can be chopped down? Should they even be handled by this block?

vhd:
I didn't plan to include tree chopping function into this machine. Even more - sap collection was implemented "just because it can be done".
IMO - "Greenhouse" is just for crops.
What do you think?

modmuss50:
I agree, I think another machine to handle trees and sap will be the best way to go about it.

src/main/java/techreborn/blockentity/machine/tier1/GreenhouseControllerBlockEntity.java
```diff
@@ -0,0 +1,241 @@
...
+			boolean breakBlocks = false;
+			for (int y = 1; (blockState = world.getBlockState(blockPos.up(y))).getBlock() == block; y++) {
+				if (y == 1) {
+					breakBlocks = tryHarvestBlock(blockState, blockPos.up(y));
+				} else {
+					tryHarvestBlock(blockState, blockPos.up(y));
+				}
+				if (breakBlocks) world.breakBlock(blockPos.up(y), false);
+			}
+		} else if (block instanceof BlockRubberLog) {
```
modmuss50, line 128: For now this can stay here, not sure if it will be better with a tree chopping block or not.

modmuss50 (approved):

PR #1965, opened 2020-01-10
Cables can now be covered

vhd:


src/main/java/techreborn/client/render/entitys/CableCoverEntityRenderer.java
```diff
@@ -0,0 +1,39 @@
...
+import net.minecraft.client.render.block.BlockRenderManager;
+import net.minecraft.client.render.block.entity.BlockEntityRenderDispatcher;
+import net.minecraft.client.render.block.entity.BlockEntityRenderer;
+import net.minecraft.client.util.math.MatrixStack;
+import techreborn.blockentity.cable.CableBlockEntity;
+import techreborn.blocks.cable.CableBlock;
+
+import java.util.Random;
+
+public class CableCoverEntityRenderer extends BlockEntityRenderer<CableBlockEntity> {
```
modmuss50, line 19: Wonder if this can be done via a FabricBakedModel? Would be more complex however

vhd: I've tried. There is lack of documentation and examples, so I give up

modmuss50 (approved):
Looks fine, might be worth looking at using the model pipeline for rendering, but like I said that will be a lot more complex than this.

Im happy to merge this as-is and improve over time if needed.

src/main/java/techreborn/client/render/entitys/CableCoverEntityRenderer.java
modmuss50: Lack of documentation? The javadocs cover everything and then some?

vhd: I constantly use decompiled versions and wasn't aware of java docs :(

PR #2015, opened 2020-01-26
Check rubber logs for sap from the ground level.

ghost:
Rubber logs can be put down manually, and it's not a given that the most bottom log is at the greenhouse's level.  In practice, with the ground made up out of rubber logs, there will be sap on logs at y == 0.


modmuss50:
thanks.

PR #2028, opened 2020-02-06
Add quartz dust recipe to grinder

JessiAuro:
The quartz dust added in 162e4bb doesn't currently have a recipe for the grinder, and can't be obtained. This PR simply adds a recipe for it copying the values from the quartz ore.


drcrazy:
Thank you for spotting!

JessiAuro:
Was it intentional that this wasn't merged?

drcrazy:
Nope. Thank you for second PR.

PR #2054, opened 2020-02-27
improved rei compat

shedaniel:
I made the arrow actually have the correct speed, marked input and output for slots for rei, etc


modmuss50:
Looks good, thanks a lot. Sorry for being slow

Issue #2046, opened 2020-02-17
Lots of useless dusts


justastranger:
**Is your feature request related to a problem? Please describe.**
Andesite, Diorite, and Granite dusts have no recipes and no uses. Diorite and Granite are ground into quartz dust now. Basalt dust has a use, but has no recipe. Flint dust has no recipe. Marble dust has a use, but once again has no recipe. Calcite can be crafted, but has no use.

**Describe the solution you'd like**
Recipes, simply put.

I'm whipping up a datapack to fill in a bunch of recipe gaps, and with the help of LibCD these recipes can be conditional based on which mods are loaded. If terrestria is loaded, I have the datapack set up to tag all their basalt blocks as "#c:basalt" (I'm surprised that the terrestria dev didn't do something similar themself...) and create a recipe that grinds them 1:1 into basalt dust. Blockus has marble, IIRC, and something similar can be set up to provide a recipe for marble dust.

[techreborn_custom_recipes.zip](https://github.com/TechReborn/TechReborn/files/4215411/techreborn_custom_recipes.zip)

None of the recipes should be overpowered and should be pretty easy to just add straight to the mod. I would've loved to have done this through a Pull Request but I don't have a development environment set up or enough data to do so.

justastranger:
The datapack also adds a use for the carbon_fiber fluid, with 2 cells being blasted into a carbon plate, being a more efficient way of making the item than with coal dust and a compressor. It can only be crafted after ashes become available.

It also adds a canning recipe for calcite -> calcium carbonate cell, so that it is analogous to how sulfur can be canned and uncanned, as well as the ability to use calcite dust instead of calcium carbonate cells in the blast furnace to make refined iron from pyrite and iron ore.

It also adds an old method of creating compressed air using a compressor and an empty cell.

The prismarine shard recipes were for the sake of uncrafting the blocks lossily as well as having a symmetry for the prismarine shard -> prismarine crystal grinder recipe.

justastranger:
Manganese, Olivine, and Phosphorous still lack any use or recipes. Thinking of making phosphorous compostable with the compostingrecipes mod. Saw Dust could be used to create charcoal like how Thermal Expansion used to do it, but I think that'd probably require at least one more item to be made.

drcrazy:
ok. I'll wait for you then

justastranger:
[...]

PR #2091, opened 2020-04-15
Blockus compatibility (1.15+)

Sturmlilie:
Building on the datapack published by @justastranger in #2046, this allows to grind basalt/marble/limestone and craft crude storage units from the different wood barrels added by blockus.

I only added the compat related changes from @justastranger because I didn't feel comfortable comitting all of them under my name.

The "vanilla" crude storage recipe should probably be hidden as well if blockus is installed as it completely replaces the plain barrels.

This is unfortunately not compile tested as I don't have a dev environment, but the recipes have all been tested via the datapack.

Edit: Removed terrestria changes

### DIFF
--- a/src/main/resources/data/techreborn/recipes/centrifuge/marble_dust.json
+++ b/src/main/resources/data/techreborn/recipes/centrifuge/marble_dust.json
@@ -4,7 +4,8 @@
   "time": 2050,
   "ingredients": [
     {
-      "item": "techreborn:marble_dust"
+      "item": "techreborn:marble_dust",
+      "count": 8
     }
   ],
   "results": [
--- a/src/main/resources/data/techreborn/recipes/crafting_table/unit/storage/crude_storage_unit_blockus.json
+++ b/src/main/resources/data/techreborn/recipes/crafting_table/unit/storage/crude_storage_unit_blockus.json
@@ -1,0 +1,29 @@
+{
+  "type": "minecraft:crafting_shaped",
+
+  "pattern": [
+    "WWW",
+    "WBW",
+    "WPW"
+  ],
+
+  "key": {
+    "B": {
+      "tag": "blockus:barrels"
+    },
+	"P": {
+      "item": "minecraft:paper"
+    },
+	"W": {
+      "tag": "minecraft:planks"
+    }
+  },
+
+  "result": {
+    "item": "techreborn:crude_storage_unit"
+  },
+
+  "conditions": {
+    "reborncore:tag": "blockus:barrels"
+  }
+}
--- a/src/main/resources/data/techreborn/recipes/grinder/marble_dust_from_limestone.json
+++ b/src/main/resources/data/techreborn/recipes/grinder/marble_dust_from_limestone.json
@@ -1,0 +1,21 @@
+{
+  "type": "techreborn:grinder",
+  "power": 2,
+  "time": 180,
+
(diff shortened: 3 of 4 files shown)


Prospector:
Adding the tags should be up to the individual mods that add the blocks. If you add a tag to a block that isn't present, the game will crash. I'm not sure how recipes that use tags with 0 values work though. (ie. basalt tag -> basalt dust recipe when there are no basalt values in the tag)

Sturmlilie:
@Prospector The libcd instructions should ensure that both the tags and recipes are only loaded when the respective mods are present, although that doesn't prevent a situation where a mod removes / changes a block ID for whatever reason.
But you're still right, the proper way to go about this is to tag the blocks within the original mods. I will make respective PRs to those.

I will leave this one open because I'm not sure whether github will allow me to reopen it had I closed it for now.

justastranger:
Yeah, when I made that datapack, the goal was "make these things work for
me", not "let's get these things done correctly" ;) so I agree, the tags
should be added in their respective mods, but the rest of the stuff should
be fine

Sturmlilie:
Aand I just found out the consensus in https://github.com/FabricMC/fabric/issues/5 was to use the `global` namespace over `c`, right as I get my blockus patch merged. Great =|  
I'll stick with `c` for this PR; if in the future `global` establishes itself among the fabric community, TR will have to be updated anyways.

Prospector:
Well that's still an ongoing debate. Currently `c` is the widely used namespace.

Sturmlilie:
I am currently in the process of splitting off the terrestria changes; they recently renamed their basalt blocks to "volcanic rock" to differentiate them from vanilla 1.16 basalt, so it's unclear whether they'll accept the tags (and if we should in the first place).  
Blockus has everything in place, however I am not happy with the conditional load of the grinder recipes. I will add a custom libcd condition that specifically checks whether the `c:marble/limestone` tags have any contents; this way it doesn't matter if a user accidentally mixes an older blockus release with a newer TechReborn one.

Sturmlilie:
This is now only related to blockus compat. One remaining hurdle: The recipes **depend** on libcd being present, otherwise grinder entities go insane (grinding nothing into nothing). So unless libcd is made a hard dependency, these changes together with any other material-related compatibility are a no-go.

modmuss50:
Our recipe system supports conditional recipes, I’m not sure if there is any documentation on it however.

modmuss50:
Oh I see you have done it using mc meta, I didn’t know that was a thing?

Sturmlilie:
@modmuss50 Where in the source tree is it?

modmuss50:
Here: https://github.com/TechReborn/RebornCore/blob/1.15/src/main/java/reborncore/common/crafting/ConditionManager.java

For items I think its like this:

```json
"conditions" : {
    "item" : "minecraft:dirt"
  }
```

`item` `fluid` `tag` or `mod` are valid atm. Im happy to add anything else that might be needed.

Sturmlilie:
@modmuss50 hmm, are there any examples of conditionally including a recipe? All uses I could find was in loot tables (as a child of the "pools" object) and in achievements (as a subchild in "criteria").
I tried adding various combinations of `"conditions": []`, `"condition":` and `"item": "minecraft:dirt"` for testing and none of them make the test recipe active.

modmuss50 (approved):

PR #2122, opened 2020-05-26
disable thing that is not tr

shedaniel:


Yoghurt4C:
This is a very nice pull request. I remember an especially difficult bit in the original code that was avoided by this hack. It has not been reported as a bug for many years, but it could be possible that it never worked correctly in the first place. Likewise, the tests don't always pass anymore, but it is yet not clear why. Also, the buffer used by minisection may be wrong (an explanation for this part would be nice) - possibly allocating extra memory or overwriting some of it. I also need to fix the last line in test, which calls the numeric (even for negative) value of each pixel. By using it without checking it (meaning it's a false positive), the negative pixel isn't captured.

PR #2124, opened 2020-05-29
Lockable storage units, WIP

### LINKED ISSUE #1973: Locking Storage Units to a certain item
**Is your feature request related to a problem? Please describe.**
After all items in a Storage Unit have been removed, it reverts to being able to accept and store any item.  Having the option to lock the Storage Unit for a certain item would make storage setups more predictable.  Even more so in chained crafting setups with multiple in- and outputs.

**Describe the solution you'd like**
The behavior of the Auto Crafting Table is pretty intuitive, and I like to see it replicated in the Storage Unit.  By adding a "lock-icon" in the interface, the player can choose whether to lock a Storage Unit for a certain item or not.  Optionally, replicating the behavior of popular barrel mods, shift-right clicking the Unit to toggle the locking.

**Describe alternatives you've considered**
A separate routing & filtering block.  Though with this added functionality to storage units, they would be able to act as both routing & filtering blocks already.

drcrazy:


src/main/java/techreborn/blockentity/storage/item/StorageUnitBaseBlockEntity.java
```diff
@@ -54,11 +58,35 @@
 public class StorageUnitBaseBlockEntity extends MachineBaseBlockEntity
 	implements InventoryProvider, IToolDrop, IListInfoProvider, IContainerProvider {
 
+	private String logPref() {
```
drcrazy: This might go to common log functionality

src/main/java/techreborn/blockentity/storage/item/StorageUnitBaseBlockEntity.java
```diff
@@ -222,6 +265,11 @@ public ItemStack processInput(ItemStack inputStack) {
 	}
 
 	public boolean isSameType(ItemStack inputStack) {
+		if (getLocked()) {
+			System.out.printf("+++ SUB#isSameType: locked: [%s]\n", lockedItem);
```
drcrazy: Just log with debug log level

Sturmlilie: Sorry, I left all this stuff in because it's not meant to go into a formal PR.

drcrazy: It is absolutely fine to have extended logging. Might help with debugging in future. Just log with debug level

Sturmlilie:
@drcrazy If you want generic debug log messages, you're going to have to tell me what specifically you want logged. My earlier logging was just scaffolding for learning how the codebase works and getting this implemented.

src/main/java/techreborn/blockentity/storage/item/StorageUnitBaseBlockEntity.java
```diff
@@ -54,11 +58,15 @@
...
 
-
 	// Inventory constants
 	private static final int INPUT_SLOT = 0;
 	private static final int OUTPUT_SLOT = 1;
 
+	// NBT keys
+	private static final String TAG_LOCKED_ITEM = "lockedItem";
+
+	private static final Item NULL_ITEM = Items.AIR;
```
Sturmlilie: I am unsure whether this isn't causing a race with the static initializer of `Items.AIR`.

modmuss50: I would use an itemstack here, as it should lock on the NBT value as well.

src/main/java/techreborn/blockentity/storage/item/StorageUnitBaseBlockEntity.java
```diff
@@ -54,11 +58,15 @@
 public class StorageUnitBaseBlockEntity extends MachineBaseBlockEntity
 	implements InventoryProvider, IToolDrop, IListInfoProvider, IContainerProvider {
 
-
 	// Inventory constants
 	private static final int INPUT_SLOT = 0;
 	private static final int OUTPUT_SLOT = 1;
 
+	// NBT keys
+	private static final String TAG_LOCKED_ITEM = "lockedItem";
```
modmuss50: Not sure this is needed here tbh, not the end of the world.

src/main/java/techreborn/blockentity/storage/item/StorageUnitBaseBlockEntity.java
```diff
@@ -68,6 +76,10 @@
 
 	private TRContent.StorageUnit type;
 
+	// A locked storage unit will continue behaving as if it contains
+	// the locked-in item, even if the stored amount drops to zero.
+	private Item lockedItem = NULL_ITEM;
```
modmuss50: Again, use an itemstack, and set it to empty.

src/main/java/techreborn/blockentity/storage/item/StorageUnitBaseBlockEntity.java
```diff
@@ -222,6 +244,10 @@ public ItemStack processInput(ItemStack inputStack) {
 	}
 
 	public boolean isSameType(ItemStack inputStack) {
+		if (isLocked()) {
+			return inputStack.getItem() == lockedItem;
```
modmuss50: When replaced with a stack use a similar check as bellow.

src/main/java/techreborn/blockentity/storage/item/StorageUnitBaseBlockEntity.java
```diff
@@ -377,13 +413,50 @@ public void onBreak(World world, PlayerEntity playerEntity, BlockPos blockPos, B
 		// Inventory gets dropped automatically
 	}
 
+	// The int methods are only for ContainerBuilder.sync()
```
modmuss50, line 407: You should be able to sync booleans?

src/main/java/techreborn/client/gui/GuiStorageUnit.java
```diff
@@ -29,9 +29,14 @@
 import reborncore.client.gui.builder.GuiBase;
 import reborncore.common.util.StringUtils;
 import techreborn.blockentity.storage.item.StorageUnitBaseBlockEntity;
+import reborncore.common.network.NetworkManager;
+import techreborn.packets.ServerboundPackets;
 
 public class GuiStorageUnit extends GuiBase<BuiltContainer> {
 
+	private static final int LOCK_BUTTON_X = 145;
+	private static final int LOCK_BUTTON_Y = 4;
```
modmuss50: Not sure we need these here, id inline them.

src/main/java/techreborn/blockentity/storage/item/StorageUnitBaseBlockEntity.java
modmuss50: Even better would be sycning the locked type so the user can see what type of item the container has been locked to.

modmuss50 (changes requested):
Looks like a good start, I would move to using ItemStack as the locking "key" would allow for NBT checking and simpler code.

src/main/java/techreborn/blockentity/storage/item/StorageUnitBaseBlockEntity.java
Sturmlilie: I have no idea how this `sync()` business works or why it's needed, I just copied it from the two other block entities that also have a lock button in their GUI (autocraft table and rolling machine). Will have to look into it.

Sturmlilie:
Changes since last version:
* Switch to ItemStack for NBT support
* Un-DRY-ify
* Push lock button further to the right to accomodate longer strings

I tried understanding the `sync()` stuff I copied from `RollingMachine` but failed to grasp it; in `ContainerBlockEntityInventoryBuilder` there is a comment about automatic syncing of the value via a custom packet, but I couldn't find where this packet exchange happens. Also if it's automatically synchronizing, why do I need to manually send / handle a packet still?

Sturmlilie:
Removed superfluous imports.

Issue #1946, opened 2020-01-03
Reborn of IC Autominer/Pump (and couple of suggestions for long distance)


Pyroksyllin:
Can we see in the future new (or old) interpretation of IC2 autominer and/or pump? Kinda corellation with [#1683](https://github.com/TechReborn/TechReborn/issues/1683), but i don't think that filling tanks with tons of buckets manually is a good idea. By this moment i don't see any ways to have, for example, lava cells. It was good way to get energy - use lava cells, pump and geothermal generator. Autominer not so necessary as pump, in my opinion, having 3*3 mining zone for top tier drill kinda compensing this. But automatisation is good.

And long run suggestions...
Can we see sometimes some adaptations of the buildcraft features? Exactly - pipes (for fluids and items) and new types of worldspawn natural resources (oil, natural gas for example)?

btw sorry for possible lack of english in my post

AndrewLischuk:
The question remains open. 

Although the Basic Tank pulls in fluids from other blocks (thanks to @modmuss50) . But it doesn't pulls in water (or lava) from open sources. Filling it manually is extremely uncomfortable. IMO this destroys the meaning of the automation process. An Industrial grinder, for example, requires a lot of water and now it's not possible to use it.
I agree with @Pyroksyllin  that Autominer is not so needed. But Pump is. Extremely. Or maybe it’s easier to make Basic Tank to pull in.

Pipes or tubes - awesome idea of course, but it's probably for the future. For now - Pump is needed.

For a few days I've been searching a lot of mods, but I didn't find a mod with Pump compatible for TR. Everything is either on Forge or on previous versions of MC.

Does anyone know how Drain will work?

THX)

Tkvgamer:
simple pipes has a pump compatible with TechReborn (https://www.curseforge.com/minecraft/mc-mods/simplepipes)

AndrewLischuk:
> simple pipes has a pump compatible with TechReborn (https://www.curseforge.com/minecraft/mc-mods/simplepipes)

Yeah, I tried this mod, but it seems doesn't work correctly on 1.15.1.... Simple Pipes's Pump filling it's tank, but ... That's it. Nothing goes on. TR's Basic Tank doesn't pulls the water from SP's Tank. Pipes doesn't work at all, or I just did something wrong... I will try ones again)))) 
Thanks for the tip, @Tkvgamer

Sturmlilie:
[...]

PR #2131, opened 2020-06-13
Drain

jusvit:
Only useful for collecting water. 
Premise is that you  put it under a 2x2 infinite water and then can pump from it (via tank unit or something else)

Recipe is on the more expensive side kinda. Just looking on approval on code (Performance-wise)

Progress towards  #1946

### DIFF
--- a/src/main/java/techreborn/blockentity/machine/misc/DrainBlockEntity.java
+++ b/src/main/java/techreborn/blockentity/machine/misc/DrainBlockEntity.java
@@ -27,22 +27,17 @@ package techreborn.blockentity.machine.misc;
 import net.minecraft.block.Block;
 import net.minecraft.block.BlockState;
 import net.minecraft.block.FluidDrainable;
-import net.minecraft.block.entity.BlockEntity;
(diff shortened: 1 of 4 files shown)


src/main/java/techreborn/blockentity/machine/misc/DrainBlockEntity.java
```diff
@@ -24,17 +24,24 @@
...
+import reborncore.common.fluid.container.FluidInstance;
+import reborncore.common.util.Tank;
 import techreborn.init.TRBlockEntities;
 
-public class DrainBlockEntity extends MachineBaseBlockEntity implements InventoryProvider, IListInfoProvider, IContainerProvider {
+import javax.annotation.Nullable;
+
+public class DrainBlockEntity extends MachineBaseBlockEntity {
+
+	protected Tank internalTank = new Tank("tank", FluidValue.BUCKET, this);
```
jusvit, line 44: Internal tank isn't persistent, shouldn't be a problem as it's not meant to be a container.

modmuss50 (approved):
looks good thanks

PR #2153, opened 2020-07-02
Create basic sulfur grinding recipes for compatibility with Cinderscapes

ghost:
While TechReborn itself doesn't add sulfur or sulfur ore, it does add sulfur dust without the necessary grinding recipes to actually acquire sulfur dust from ores added by other mods. This PR adds two simple recipes that use the c:sulfurs tag and the c:sulfur_ores tag to grind into their respective quantities of sulfur_dust.

Specifically this PR will resolve compatibility with Cinderscapes which adds both a sulfur item and sulfur ore that should be able to be placed in the grinder so it can be used with TechReborn. It doesn't really make sense to keep these recipes in Cinderscapes itself as it doesn't relate to tech whatsoever, or add any sulfur dust of its own.

### DIFF
--- a/src/main/resources/data/techreborn/recipes/grinder/sulfur_dust.json
+++ b/src/main/resources/data/techreborn/recipes/grinder/sulfur_dust.json
@@ -1,0 +1,15 @@
+{
+  "type": "techreborn:grinder",
+  "power": 2,
+  "time": 200,
+  "ingredients": [
+    {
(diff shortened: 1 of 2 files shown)


jusvit:
Look alright, not much use for sulfur in TR besides refining, not exactly sure how rare sulfur ore is in Cinderscapes but it shouldn't have a significant impact on balance.

PR #2168, opened 2020-07-19
Create soul_soil.json

ghost:
Copy of the centrifuge soul_sand recipe, but for soul_soil
I'm sorry, I don't know if I need to add a reference to this recipe anywhere else.


Prospector:
If the recipe is identical, then it might be a better idea to make a tag for soul sand and soul soil, and use the tag in the recipe.

ghost:
I've adjusted the recipe to return 8 minecraft:dirt instead of 10 minecraft:sand.

jusvit:
Alright when the dirt changes which makes more sense, everything else looks good.

PR #2199, opened 2020-08-17
Added back initial Applied Energistics 2 compatibility

shartte:
I suppose you might not want this in your main repo, But the compatibility repo seemed to be abandoned since the recipe changes, so I'll PR this here and you can move it where it needs to be.

The values are all based off the old compatibility code from 1.12.


modmuss50 (approved):

modmuss50:
Yeah, the mod compat repo is a 1.12, no real need for it in 1.16 with json recipes

Issue #2127, opened 2020-06-03
Solar panel energy generation should scale with light incidence angle


Sturmlilie:
**Is your feature request related to a problem? Please describe.**
I think everyone who has played TR will agree that solar panels are extremely OP and just blow every other generator out of the water; this is a bit sad because setting up and managing an energy generation process (eg. from natural / grown resources) can be quite fun. Solar panels OTOH just sit there.

**Describe the solution you'd like**
The simplest iteration of this feature I could think of is: Make energy production during the day proportional to the sine of the light incidence angle, with the angle being 90° during noon and 0° during sunrise/set. For consistency, I'm suggesting the minimum of this to be capped to the night generation rates, so the E/t don't go down to zero during sunset and jump back up again when night falls (I'm interpreting night generation as "ambient light generation" here).  
Without tweaking the existing day/night generation rates, this is already a nerf in two ways:
* Energy generation is no longer consistent and must be buffered where necessary
* The total amount of energy produced during one day is 2 over pi, or roughly 60% times that of the current flat generation

In my opinion, solar panels are still OP even with this change, simply due to their high raw output, but this can be tweaked via the config; maybe a consensus on more sensible default values can be reached.

Edit: On fabricord it has been suggested to make this feature optional via config.

**Describe alternatives you've considered**
Just tweaking the default day generation rates works, but I think this feature can make panels actually more interesting.

**Additional context**
I've made a prototype of my suggestion which you can try out:  
[**[jar link]**](https://srv-file14.gofile.io/download/vwDbCI/TechReborn-1.15-3.3.11+build.local.jar)  
[**[source code]**](https://github.com/Sturmlilie/TechReborn/tree/solar_angle)  
[**[video demonstration]**](https://youtu.be/kcd8kq8pFkU) (daylight cycle sped up by 60x)

jusvit:
I mean it isn't a bad idea, I like the uniqueness, wouldn't just throw it out. Don't see any reasons it couldn't be something you could enable in config, and allow modpack creators to balance.

Sturmlilie:
I'm not throwing it out, just not sure I wanna spend more time on TR right now. The idea and my mockup code are still there, I won't delete the branch. If there is some actual input / consent on the game design aspect of it I'll probably finnish it though.

Technici4n:
Implemented in #2497.

Sturmlilie:
Rip sinusoidal output :(

PR #2213, opened 2020-09-06
Implement sun incidence angle scaling for solar panels

Sturmlilie:
image

Details can be found in the accompanying issue: #2127 .

Right now, solar panels are incredibly overpowered, pushing all other generator types into oblivion. This new feature will not entirely remedy this, and I feel that either solar panels need to be nerfed even further (by at least halving their output), or by generously buffing the fuel- and other regenerative-energy based power sources.

Regarding the implementation, I feel a bit conflicted about the hack I had to build in using the counter `int` that is increased every time the to-be-displayed contents (in this case, light angle and energy output) change, so the GUI is updated accordingly. I'm hoping for suggestions on how to do that in a less hacky way.

Also remember that despite this looking a bit dicey for a ticking entity, the angle calculations are only done every few ticks (20 I believe).

I have tested this successfully on dedicated server.

### DIFF
--- a/src/main/java/techreborn/blockentity/generator/SolarPanelBlockEntity.java
+++ b/src/main/java/techreborn/blockentity/generator/SolarPanelBlockEntity.java
@@ -34,6 +34,7 @@ import net.minecraft.text.Text;
 import net.minecraft.text.TranslatableText;
 import net.minecraft.util.Formatting;
 import net.minecraft.util.math.Direction;
+import net.minecraft.world.World;
 import reborncore.api.IToolDrop;
 import reborncore.client.screen.BuiltScreenHandlerProvider;
 import reborncore.client.screen.builder.BuiltScreenHandler;
@@ -48,6 +49,7 @@ import techreborn.init.TRBlockEntities;
 import techreborn.init.TRContent;
 import techreborn.init.TRContent.SolarPanels;
 
+import java.lang.Math;
 import java.util.List;
 
 public class SolarPanelBlockEntity extends PowerAcceptorBlockEntity implements IToolDrop, BuiltScreenHandlerProvider {
@@ -63,6 +65,10 @@ public class SolarPanelBlockEntity extends PowerAcceptorBlockEntity implements I
 	private int state = ZEROGEN;
 	private int prevState = ZEROGEN;
 
+	public double incidenceAngle = 0;
+	private int generationRateD = 0;
+	private int updateCounter = 0;
+
 	private SolarPanels panel;
 
 	public SolarPanelBlockEntity() {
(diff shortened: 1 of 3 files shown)


jusvit:
Could you add a config option and modifications for this to be disabled/enabled for modpack creators/users?

Sturmlilie:
Yeah, sure. Default enabled though, right?

jusvit:
> Yeah, sure. Default enabled though, right?

Kinda conflicted, I think just do enabled for now and I can change it, need to balance it out but also don't want to ruin half the player's power system unless it's a major new MC version.

Sturmlilie:
Is it possible that the values in `TechRebornConfig` aren't synced server->client at connection time? Because if not, this is going to get really annoying. I don't wanna have to send update packets for each energy generation change.

Sturmlilie:
Actually, if you're concerned about smashing people's solar farms in the middle of a stable version, maybe scratch the config setting and just merge this once the 1.17 snapshot cycle starts? I'll make sure that it still builds then.

modmuss50:
If you sync the values only in the gui then there will be no packets sent unless the gui is open

Sturmlilie:
@modmuss50 Alright I'll try that, thanks (though it's still really annoying because we have all the data we need to correctly calculate all the values client-side, we just need to know the config the server is running off of).

jusvit:
> @modmuss50 Alright I'll try that, thanks (though it's still really annoying because we have all the data we need to correctly calculate all the values client-side, we just need to know the config the server is running off of).

The calculations should all be server-side. The ServerWorld should have all the fields you'll require for the incidence angles.

Sturmlilie:
@justinvvitale Yes, the server is the authority of truth and thus also has all parameters needed to arrive at the resulting energy output. The point is, server and client can agree on the correct energy output without having to communicate at all, since they both have their sky rotation values synced anyway. The initial patchset by me only ever sends fake packets to force the GUI to update. Though one might argue, if I'm already sending a dummy integer to tell the UI to update, I might as well send the correct energy output using that pipe.

modmuss50 (approved):

Sturmlilie:
@modmuss50 I'm assuming that "approval" was regarding my last proposal for a change? If so, I'll implement that sometime Tuesday or Wednesday when I'm back home.

Sturmlilie:
I have hit a roadblock. I am realizing that my fundamental misunderstanding of this stuff:
image
is preventing me from progression.

**What I thought it did**: Once a GUI is opened, continuously call the getter on the server, transmit this value over the network, and invoke the setter on the client with the transmitted value.

**What actually happens**: Both getter and setter are just executed on the client once the GUI is opened.

Technici4n:
Replaced by #2497.

PR #2238, opened 2020-10-02
Fix several static property fields being reassigned when block instances call appendProperties

Virtuoel:


modmuss50 (approved):
Great, thanks for spotting and fixing this

PR #2260, opened 2020-10-26
Update en_us.json

Romz24:
I found that some keys are missing.

image


Romz24:
If approved, please do not forget to add new keys to translate.techreborn.ovh

PR #2338, opened 2021-02-13
Fix the solar panel overfill bug

triangular2357:


triangular2357:
please don't merge it yet

triangular2357:
I will make a new pr with a better solution

PR #2342, opened 2021-02-14
21w11a

ghost:
screenshot2

**TODO:**

- [ ] Figure out what to do with copper
- [x] Figure out how to get a Fluid from FluidDrainable.tryDrainFluid (it now returns ItemStack)
- [ ] Add Rubber logs to the correct Tag


src/main/java/techreborn/blockentity/generator/basic/WindMillBlockEntity.java
```diff
@@ -42,13 +46,13 @@
...
-	public WindMillBlockEntity() {
-		super(TRBlockEntities.WIND_MILL);
+	public WindMillBlockEntity(BlockPos pos, BlockState state) {
+		super(TRBlockEntities.WIND_MILL, pos, state);
 	}
 
 	@Override
-	public void tick() {
-		super.tick();
+	public void tick(World world, BlockPos pos, BlockState state, MachineBaseBlockEntity blockEntity) {
```
modmuss50, line 54: Are these tick methods not supposed to be static? Its been a while since I last looked at it so id have to check the code. Vanilla has a client and server one if im not mistaken, and your supposed to modify the data on the BE passed in as a param.

ghost: The way vanilla does it is completely without an interface. Blocks have a `getTicker(...)` function now and that returns the static tick function.
I don't know about a client and a server one though.
I went for this solution (implementing BlockEntityTicker on the BlockEntity), so we can have one generic `getTicker(...)` in BaseBlockEntityProvider that calls the respective (non-static) `tick()` function of the BlockEntity. I think it should work just fine this way and it saves us the hassle of writing a getTicker function for each Block and also registering different tick-functions for BlockEntities that use a shared Block class like GenericMachineBlock.

modmuss50:
As for the copper i'd love to find a way to migrate people's worlds, I dont think its going to be super easy to do though.

ghost:
I could imagine something similar to the old RetroGen mods: When loading chunks for the first time, it replaces all TechReborn copper blocks with vanilla ones and marks the chunk as 'processed'. This would mean there would still be a TechReborn copper block. It would just be hidden from the creative tab, not generated, ....
For ingots etc. it's a bit more difficult, but it would be possible to convert them to vanilla items when placed in the player inventory, like some older modpacks did for all ingots that were not from the mod that's supposed to provide this kind of Item.
Not really seamless, but better than nothing. A nicer solution would be to be able to remove the copper items and blocks entirely and somehow convert everything to the vanilla variants when reading the world data, but I imagine that would be really hard.

ghost:
Ouch: The vanilla Copper Block gives only 4 Copper Ingots :frowning_face:. This will be _fun_.

modmuss50:
Another way would be to just keep the exiting copper blocks and items there. Hide them from the create tab and REI and add a crafting recipe to convert them.

It’s going to be a bit of a pain to rebalance the recipes to deal with the change of size shouldn’t be too awful I hope.

drcrazy:
Keep both TR and vanilla. Recipe rebalance could be done closer to 1.17 release and shouldn't block overall update to 1.17

Also I'm in doubt that any 1.16 modded world could be migrated to 1.17

Technici4n:
Why worry about world migration from 1.16 to 1.17? I can tell you that 1.16 worlds with Modern Industrialization will not work in 1.17 anyway because that will be the occasion to change the serialization format and cleanup bad IDs. Not to mention mods that won't even update at all.

modmuss50:
We could quite easily get away without supporting upgrade, I dount many people will try. However its fairly simple to do something basic like adding crafting recipes to convert to the new items we might as well do it.

ghost:
21w07a starts, but no ores are generated

ghost:
Some friends have been playing the 21w08b commit(s) for about two weeks now, and they haven't noticed any big bugs.

modmuss50:
Glad I checked here as I honestly forgot about this. I have started a 1.17 branch of TR that now includes RC as a sub project. I will get your 2 PRs merged into that manually. This seems to cover most of the major items that require porting 👍

modmuss50:
I have manually merged this into the 1.17 branch running on pre1, many thanks for this!

PR #2367, opened 2021-03-11
Use tags in steel and refined iron blast furnace recipes

saltyseadoggo:
This pull request makes the blast furnace's steel from refined iron and coal dust recipe use `#c:coal_dusts` for compatibility with coals dusts from other mods like Astromine, and also makes its refined iron from iron ore recipe use `#c:iron_ores` for compatibility with modded iron ore variants, like those from UnEarthed.


drcrazy:
Thanks!

PR #2372, opened 2021-03-19
Fix CTM compatibility

YuRaNnNzZZ:
Fixed compatibility with [CTM for Fabric](https://www.curseforge.com/minecraft/mc-mods/ctm-fabric).

JSON model override for casings is used to ensure proper texture is used regardless if CTM is installed or not.


drcrazy:
Awesome! Thanks you!

PR #2386, opened 2021-04-14
Fix ConcurrentModificationException due to REI async search

Fourmisain:
## Description

This is more of a band-aid solution to a slightly bigger problem:
`ItemTooltipCallback.getTooltip()` can be executed by worker threads when REI is in async search mode and it is not completely thread-safe.

Going through it:
- `ITEM_ID` is not thread-safe and `computeIfAbsent` can cause a `ConcurrentModificationException` (see the log below)
- `isTRItem` uses `Registry.ITEM` which is not thread-safe (uses `BiMap`), this shouldn't be a big issue since the registry doesn't change after init and it can't throw
- `item.getTranslationKey()`: can use `Registry.ITEM` too
- `ToolTipAssistUtils.getUpgradeStats`: accesses `TechRebornConfig` which is not thread-safe, but also unlikely to change

So except `ITEM_ID`, it should work well enough as-is - also, there's little one can do to make access to `Registry.ITEM` thread-safe.

## Changes

- Use a `ConcurrentHashMap` for `ITEM_ID`
- make `isTRItem` `static`, simply because it can be `static`

### Log (formatted to fit)

> \[ForkJoinPool.commonPool-worker-11/INFO]:
> java.util.ConcurrentModificationException
 at java.base/java.util.HashMap.computeIfAbsent(HashMap.java:1225)
 at techreborn.events.StackToolTipHandler.getTooltip(StackToolTipHandler.java:56)
 at net.fabricmc.fabric.api.client.item.v1.ItemTooltipCallback.lambda$static$0(ItemTooltip
 at net.minecraft.class_1799.handler$cli001$getTooltip(class_1799.java:4155)
 at net.minecraft.class_1799.method_7950(class_1799.java:767)
 at me.shedaniel.rei.impl.ItemEntryStack.tryGetItemStackToolTip(ItemEntryStack.java:400)
 at me.shedaniel.rei.impl.ItemEntryStack.getTooltip(ItemEntryStack.java:289)
 at me.shedaniel.rei.impl.SearchArgument.tryGetEntryStackTooltip(SearchArgument.java:180)
 at me.shedaniel.rei.impl.search.TooltipArgument.matches(TooltipArgument.java:71)
 at me.shedaniel.rei.impl.search.TooltipArgument.matches(TooltipArgument.java:42)
 at me.shedaniel.rei.impl.SearchArgument.matches(SearchArgument.java:176)
 at me.shedaniel.rei.impl.SearchArgument.canSearchTermsBeAppliedTo(SearchArgument.java:164
 at me.shedaniel.rei.gui.widget.EntryListWidget.canLastSearchTermsBeAppliedTo(EntryListWid
 at me.shedaniel.rei.gui.widget.EntryListWidget.lambda$updateSearch$6(EntryListWidget.java
 at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.jav
 at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.exec(CompletableFuture.ja
 at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:295)
 at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1
 at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1665)
 at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1598)
 at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:183)

### DIFF
--- a/src/main/java/techreborn/events/StackToolTipHandler.java
+++ b/src/main/java/techreborn/events/StackToolTipHandler.java
@@ -27,6 +27,7 @@ package techreborn.events;
 import com.google.common.collect.Maps;
 import net.fabricmc.fabric.api.client.item.v1.ItemTooltipCallback;
 import net.minecraft.block.Block;
+import net.minecraft.client.MinecraftClient;
 import net.minecraft.client.gui.screen.Screen;
 import net.minecraft.client.item.TooltipContext;
 import net.minecraft.item.Item;
@@ -34,6 +35,7 @@ import net.minecraft.item.ItemStack;
 import net.minecraft.text.Text;
 import net.minecraft.util.registry.Registry;
 import reborncore.common.BaseBlockEntityProvider;
+import techreborn.TechReborn;
 import techreborn.init.TRContent;
 import techreborn.items.UpgradeItem;
 import techreborn.utils.ToolTipAssistUtils;
@@ -53,7 +55,12 @@ public class StackToolTipHandler implements ItemTooltipCallback {
 	public void getTooltip(ItemStack stack, TooltipContext tooltipContext, List<Text> tooltipLines) {
 		Item item = stack.getItem();
 
-		if (!ITEM_ID.computeIfAbsent(item, this::isTRItem))
+		// Can currently be executed by a ForkJoinPool.commonPool-worker when REI is in async search mode
+		// We skip this method until a thread-safe solution is in place
+		if (!MinecraftClient.getInstance().isOnThread())
+			return;
+
+		if (!ITEM_ID.computeIfAbsent(item, StackToolTipHandler::isTRItem))
 			return;
 
 		// Machine info and upgrades helper section
@@ -71,7 +78,7 @@ public class StackToolTipHandler implements ItemTooltipCallback {
 		}
 	}
 
-	private boolean isTRItem(Item item) {
+	private static boolean isTRItem(Item item) {
 		return Registry.ITEM.getId(item).getNamespace().equals("techreborn");
 	}
 }


modmuss50:
None of this is thread safe, this also calls down into Screen.hasShiftDown I hightly doubt that is thread safe.

IMO this is a bug in REI, threaded search should be opt in for items, with a propper api.

Fourmisain:
> this also calls down into Screen.hasShiftDown I hightly doubt that is thread safe.

This uses `GLFW.glfwGetKey`, which according to the [docs ](https://www.glfw.org/docs/3.3/group__input.html)"may be called from any thread."

I agree that this is more of an issue with REI than anything though.

modmuss50:
```java
public CompletableFuture<List<Text>> getTooltip(ItemStack stack, TooltipContext tooltipContext)
```

might do the job from REI's side. Passing in a copy of the item stack, id have to have the code in front of me to see what TooltipContext contains.

Is this threaded searching on by default? Is there a bug report open with REI? At the very least id just put `synchronized` on the getTooltip method.

Fourmisain:
Oh, hold on, you're right, `glfwGetKey` is the one taking a handle and it says
> This function must only be called from the main thread.

modmuss50:
In that case then, I would add a check somewhere and fail hard to ensure that this is only ever ran on the main thread better than possibly getting into a weird state.

With a proper API from REI that allows extra data to be provided for search (doesnt even need to be based around the tooltip) that seems a much better solution imo.

Fourmisain:
> Is this threaded searching on by default? Is there a bug report open with REI?

Just tested to make sure and yes, it's on by default and there's no open issue I can find.

> In that case then, I would add a check somewhere and fail hard to ensure that this is only ever ran on the main thread better than possibly getting into a weird state.

Instead of hard failing we could make it a NOP, I'd rather have a running game than not being able to search for TR tooltips.
Hm... maybe we can just filter by thread name, only allowing `main`?

modmuss50:
That’s a better idea. There is a method somewhere in mc that checks if it’s the main thread.

Fourmisain:
> There is a method somewhere in mc that checks if it’s the main thread.

Looks like `MinecraftClient.getInstance().isOnThread()` is the one.

Fourmisain:
I'm gonna open an issue on REI.

modmuss50 (approved):

PR #2390, opened 2021-04-23
Fix relative path resolution in build script

Fourmisain:
This is a really small build fix.
`new File(relativePath)` (or `new File(relativePath, child)`) is (apparently) not guaranteed to be based off of the project directory and thus the `compileRecipes` task can fail.

I'm on a Windows machine and building via `gradlew build` or using IntelliJ IDEA 2020.3.3 works just fine.
Using IDEA 2021.1 however breaks the build, since it bases the path off of `%userprofile%\.gradle\daemon\6.8.3` for some unknown reason, which could probably be considered a bug - still, it shouldn't hurt fixing the build itself.

I think it could also be fixed by prepending the `recipePath` with `${projectDir}` since [`file()`](https://docs.gradle.org/current/javadoc/org/gradle/api/Project.html#file-java.lang.Object-) "Resolves a file path relative to the project directory of this project".

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -311,7 +311,7 @@ task compileRecipes {
 	def smeltingPath = recipePath + "/smelting"
 	def blastingPath = recipePath + "/blasting"
 
-    doLast {
+	doLast {
 		// Add any recipe filenames that don't match the regex here
 		String[] whitelist = []
 		def pattern = ~'^\\w+_(ingot(_from_dust)?)\\.json$'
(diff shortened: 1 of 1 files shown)


modmuss50 (approved):

Fourmisain:
Just for anybody that might stumble upon this: this [IDEA issue](https://youtrack.jetbrains.com/issue/IDEA-265203) has now been fixed in 2021.1.1.

PR #2401, opened 2021-05-24
fix recipes

triangular2357:
fixes a clay dupe and a duplicated recipe


jusvit:
Look's good, good catch.

PR #2420, opened 2021-06-25
Greenhouse Controller try harvest/grow blocks in order instead of at random + growth fix

trobol:
The randomness of the greenhouse makes it inconsistent. This patch makes it attempt to harvest/grow each crop after a consistent interval by iterating over each block in order. 

Also I'm not 100% sure, but I think that the growth acceleration wasn't working so I fixed that.

### DIFF
--- a/src/main/java/techreborn/blockentity/machine/tier1/GreenhouseControllerBlockEntity.java
+++ b/src/main/java/techreborn/blockentity/machine/tier1/GreenhouseControllerBlockEntity.java
@@ -60,6 +60,10 @@ public class GreenhouseControllerBlockEntity extends PowerAcceptorBlockEntity
 	private BlockPos multiblockCenter;
 	private int ticksToNextMultiblockCheck = 0;
 	private boolean growthBoost = false;
+	private int workingIndex = 0;
+	// number of blocks from center
+	private int range = 4;
+
 
 	public GreenhouseControllerBlockEntity(BlockPos pos, BlockState state) {
 		super(TRBlockEntities.GREENHOUSE_CONTROLLER, pos, state);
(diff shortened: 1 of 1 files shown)


src/main/java/techreborn/blockentity/machine/tier1/GreenhouseControllerBlockEntity.java
```diff
@@ -81,7 +92,7 @@ private void workCycle() {
 			) {
 				if (getStored(EnergySide.UNKNOWN) > TechRebornConfig.greenhouseControllerEnergyPerBonemeal) {
 					useEnergy(TechRebornConfig.greenhouseControllerEnergyPerBonemeal);
-					blockState.scheduledTick((ServerWorld) world, blockPos, world.random);
+					blockState.randomTick((ServerWorld) world, blockPos, world.random);
```
drcrazy, line 95: These are same
image

trobol: My understanding is that is the default behavior. I think CropBlock (and similar types) override that. Also in my testing there was only noticeable growth with `randomTick`
(From CropBlock.java)
image

trobol: For instance this is for Cactus:
image

drcrazy: Yeah, sorry, I haven't checked crop blocks

trobol: np, thanks for accepting the PR!

PR #2426, opened 2021-07-03
Fix that the mining speed of drills too slow.

### LINKED ISSUE #2425: Mining Speed of Drills is too slow for Ores
**Describe the bug**
When you remove [Ores](https://wiki.techreborn.ovh/doku.php?id=world_generation:ore) added by Tech Reborn (Tin, Lead, Silver, ...) with [**Drills**](https://wiki.techreborn.ovh/doku.php?id=items:basic_drill&redirect=1), the mining speed is slower than normal pickaxes such as Iron, Diamond and Ruby.

**Steps to Reproduce**
- Place Ores.
- Dig with Drill.
- Very slow!

 - Minecraft: 1.17
 - Mod Loader: fabric
   - fabric-loader:0.11.3
 - TechReborn-5.0.3-beta+build.55

**Additional context**
I found which code should be fixed, so I will create PR.

Kotori316:
Fix #2425

### DIFF
--- a/src/main/java/techreborn/items/tool/DrillItem.java
+++ b/src/main/java/techreborn/items/tool/DrillItem.java
@@ -86,18 +86,20 @@ public class DrillItem extends PickaxeItem implements EnergyHolder, ItemDurabili
 			return false;
 		}
 
-		if (Items.DIAMOND_PICKAXE.isSuitableFor(blockIn)) {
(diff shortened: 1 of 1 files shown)


drcrazy:
This will not fix breaking speed for pickaxes from vanilla and other mods. Tags are better solution

drcrazy:
Big thanks for suggested fix, anyway!

PR #2474, opened 2021-08-16
Initial Energy API refactor

Technici4n:
Not totally tested, and there are issues with cables not rendering the correct visual connections, but most things appear to be working (I tested tools, solar panels, cables, crafting machines). There will certainly be more bugs.

A few comments:
- `PowerAcceptorBlockEntity` was reworked a little to use `SimpleSidedEnergyContainer` internally, but still delegate to its overrideable methods.
- Most uses of `Energy.valid` and `Energy.of` were just to easily interact with the contained energy of `ItemStack`s. This was moved to `SimpleBatteryItem` in the energy API as it can be quite useful for other mods I think.
- `EnergyTier` was replaced by `RcEnergyTier`, it's just used internally to ensure consistent I/O limits.
- Imports probably need a cleanup pass.


RebornCore/src/main/java/reborncore/common/recipes/IUpgradeHandler.java
```diff
@@ -39,7 +39,7 @@
 
 	double getPowerMultiplier();
 
-	double getEuPerTick(double baseEu);
+	long getEuPerTick(long baseEu);
```
modmuss50, line 42: Not really something to worry too much here, still funny to see "EU" refrences around the place. At somepoint we should clean this stuff up.

PR #2493, opened 2021-09-02
Network energy cables

Technici4n:
Energy cable networks are now much smarter: they tick all at once and try to evenly distribute energy across all of the connected targets.


dizzyd:
FWIW, validated that this works well in both single and multi-player setups. The unified cable tick is nifty. :)

PR #2495, opened 2021-09-13
Enable overfill checking on fusion reactor energy stored

dizzyd:
It's not clear why the overfill checking was disabled on the Fusion Reactor, but this PR re-enables it so that the reactor never overfills on energy.

(This may have been a deliberate design decision - if so, feel free to close this PR out with extreme prejudice.) :)


Technici4n (approved):
It was disabled in a commit 3 years ago, and I can't find the reason, so I assume this fix is appropriate.

PR #2500, opened 2021-09-16
Buff solar panels

Technici4n:
Meant to go in alongside #2497 which introduces a solar panel nerf.

My idea is to make quantum solar panels the most effective, that's why they have a peak rate of 512 E/t while the previous tier (ultimate) only has a peak output of 48 E/t.


Technici4n:
Opened from the feature branch instead of my private branch... 😕

PR #2507, opened 2021-09-19
Fix 2504: Don't check energy capacity when reading NBT, but check after applying upgrades

Technici4n:
The added check after upgrades ensures that energy gets deleted if an upgrade is removed.


RebornCore/src/main/java/reborncore/common/powerSystem/PowerAcceptorBlockEntity.java
```diff
@@ -41,6 +41,7 @@
 import net.minecraft.util.math.Direction;
 import net.minecraft.util.math.MathHelper;
 import net.minecraft.world.World;
+import org.jetbrains.annotations.ApiStatus;
```
Technici4n: How did this get there...

PR #2510, opened 2021-09-19
Fix 2509; reviewed all ticking entities to ensure consistency in checks

dizzyd:
We've had a couple bugs pop up lately where it was the result of trying to tick on the client side. This patch goes through all the ticking entities (that I could find) and ensures the ordering of calling super.tick and the client-side checks are reasonably consistent.


kwpugh:
Wonderful!

Technici4n (approved):

PR #2573, opened 2021-10-13
Update GuiBuilder.java

PTOM76:
I don't think Fabric has JEI.

- ps
excuse me.
But will this method be used in the future?


modmuss50:
Does this button actually work? Is it even used?

The method should also be renamed if so.

PR #2585, opened 2021-10-28
Fix #2574: Use the fabric item API for auto input and auto output

### LINKED ISSUE #2574: Move to Fabric's item transfer API
This only requires updating the auto-input and auto-output code I think? This will reduce the amount of code we need for inventory interactions, and ensure compatibility with all other fabric mods that use the transfer API (as they should).

Technici4n:


RebornCore/src/main/java/reborncore/common/blockentity/SlotConfiguration.java
```diff
@@ -291,53 +289,14 @@ private void handleItemInput(MachineBaseBlockEntity machineBase) {
...
-					inventory.getStack(slotID).increment(extractedStack.getCount());
-				}
-				inventory.setHashChanged();
-				break;
-			}
+			StorageUtil.move(
+					ItemStorage.SIDED.find(machineBase.getWorld(), machineBase.getPos().offset(side), side.getOpposite()),
+					InventoryStorage.of(machineBase, null).getSlot(slotID),
+					iv -> true,
+					Long.MAX_VALUE,
```
modmuss50: This was 4 before, I think this should be kept the same speed?

Technici4n: Fixed.

modmuss50 (approved):

PR #2593, opened 2021-11-03
Fix missing hitbox of single block cable

nalepamarcin:
It prevented user from destroying last cable piece as it could not
be hit without being connected to other cable or a machine.

Bug was introduced in commit f39c619c "Fix cable shape computation".


Technici4n:
image

Damn it! Thanks a lot! :smile:

PR #2631, opened 2021-12-06
Use custom tool materials

Technici4n:
~~Need to check if it's useful first, this is untested atm.~~ Works.


Technici4n:
going to wait as there might still be a few issues

modmuss50 (approved):

PR #2646, opened 2021-12-12
Fix chainsaws don't work for non-wood materials

### LINKED ISSUE #2637: [1.17.1] Chainsaws don't work well with pumpkin
**Describe the bug**
Breaking a pumpkin takes the same time with bare hands when using chainsaws.
EDIT: as @xanthian says, it works fine on trees.

**Steps to Reproduce**
Break a pumpkin with chainsaws.

 - Minecraft: 1.17.1
 - Mod Loader: fabric
 - Tech Reborn: `TechReborn-5.0.11-beta.jar`

equal-l2:
Fixes #2637.

---

I'm totally a newbie at modding, so feel free to give me any advice!

### DIFF
--- a/src/main/java/techreborn/items/tool/ChainsawItem.java
+++ b/src/main/java/techreborn/items/tool/ChainsawItem.java
@@ -25,7 +25,6 @@
 package techreborn.items.tool;
 
 import net.minecraft.block.BlockState;
-import net.minecraft.block.Material;
 import net.minecraft.enchantment.EnchantmentHelper;
 import net.minecraft.enchantment.Enchantments;
 import net.minecraft.entity.LivingEntity;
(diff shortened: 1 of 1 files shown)


xanthian (commented):
Rather than removing the existing material check, would it not have been better to add the pumpkin one to it?

state.getMaterial() == Material.GOURD

This would include Carved and Melon

equal-l2:
Adding materials is fine, but is there a reason to list all suitable materials by ourselves, instead of delegating the decision to the vanilla logic?
People would assume chainsaws should work the same way with vanilla axes.

equal-l2:
I think my solution is more future-proof for when Minecraft add new material that is suitable for axes.

xanthian:
just did a test build with both, they both work fine, you're right though best to let vanilla declare it unless anything needs declaring from another mod.

One thing to note, cant carve pumpkins with it.

equal-l2:
> One thing to note, cant carve pumpkins with it.

Thank you for letting me know, it would deserve another issue.

equal-l2:
Wait, carving pumpkins is a job for shears, isn't it?

equal-l2:
Thank you for rapid reviewing and merging!

drcrazy:
This one is merged. Using reference tool is a good idea.

xanthian:
> Wait, carving pumpkins is a job for shears, isn't it?

yes lol.. no idea what i was thinking.. maybe carving with chainsaws isn't such a bad idea 😂

equal-l2:
I remember those days with IC2, when we sheared sheep with a chainsaw :)

PR #2651, opened 2021-12-16
Simplified centrifuging meat with tags

Ayutac:


Ayutac:
Oh, new commits in the meantime get added automatically, good to know <_<

PR #2655, opened 2021-12-18
Blast Furnace additional recipes & partial returns for damaged items

LauchInterceptor:
This PR adds Blast Furnace Recipes for smelting horse armor and other tools back into thier respective crafting ingredients.
Also it implements a mechanic for diminished returns when attempting to smelt damaged items.

Since this is my first contribution and I am fairly new to modding and this mod in particular, I am open to critique and would be very happy to have someone review this even if it in the end does not end up in the mod.


src/main/java/techreborn/blockentity/machine/multiblock/BlastFurnaceRecipeCrafter.java
```diff
@@ -0,0 +1,114 @@
...
+			// If it doesn't have all the inputs reset
+			if (isInvDirty() && !hasAllInputs()) {
+				currentRecipe = null;
+				currentTickTime = 0;
+				setIsActive();
+			}
+			// If it has reached the recipe tick time
+			if (currentRecipe != null && currentTickTime >= currentNeededTicks && hasAllInputs()) {
+				boolean canGiveInvAll = true;
+				List<ItemStack> recipeOutput = getPartialOutput(currentRecipe, getCurrentInputs());
```
drcrazy: Wouldn't it be better to override getOutputs method of BlastFurnaceRecipe instead of additional crafter?

LauchInterceptor: I agree. I tried that first but i could not find a way to get the input itemstack damage from that context. I will try again. Thank you

LauchInterceptor: Upon further investigation. I sadly could not find a nice way to determine the amount of returned items in the getOutputs method.

PR #2658, opened 2021-12-19
Fixed activation of root advancement.

Ayutac:
Fixed activation of root advancement.

### DIFF
--- a/src/main/java/techreborn/init/TRContent.java
+++ b/src/main/java/techreborn/init/TRContent.java
@@ -374,6 +374,8 @@ public class TRContent {
 	private final static Map<Ores, Ores> deepslateMap = new HashMap<>();
 
 	public enum Ores implements ItemConvertible {
+		// when changing ores also change data/techreborn/tags/items/ores.json for correct root advancement display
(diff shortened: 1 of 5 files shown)


Technici4n (changes requested):
The tag changes should be done in datagen, not manually.

modmuss50:
The data gen setup needs redoing, for now just continue with the jsons.

Ayutac:
> The tag changes should be done in datagen, not manually.

Then that should be commented instead once the redoing is done. It's important people know when changes in a certain area imply changes needed elsewhere.

modmuss50 (approved):
I wouldn’t worry about data gen too much now.

Technici4n:
> It's important people know when changes in a certain area imply changes needed elsewhere.

The whole point of moving to datagen is so that the changes only need to be done in one place...

Ayutac:
> The whole point of moving to datagen is so that the changes only need to be done in one place...

Is there a wiki page or readme in the code that explains how exactly that is done? I'm new to modding and very happy I am able to understand the code so far; I feel future contributors would be confused about the data gen as well.

PR #2659, opened 2021-12-19
Added Silver Armor. Textures by Xanthian.

Ayutac:
Silver is 2 times as durable as gold (still less than iron), -5 of gold's enchantability (still more than diamond and netherite) and cannot pacify/distract piglins. Also, it looks very fancy.


drcrazy:
Can you update this page also, please?  https://wiki.techreborn.ovh/doku.php?id=items

Ayutac:
will do EDIT: done

PR #2662, opened 2021-12-19
Fixed Deepslate Galena Ore. Texture by Xanthian.

Ayutac:


drcrazy:
Let this one stay here until next release. We've got deepslate galena ore texture updated by author of original textures.

drcrazy:
We've got texture from Spearkiller.

PR #2665, opened 2021-12-19
Capacity increase for MFE and MFSU

LauchInterceptor:
Increased the stored energy values to be more in line with the amount of resources going into the recipe

### DIFF
--- a/src/main/java/techreborn/blockentity/storage/energy/HighVoltageSUBlockEntity.java
+++ b/src/main/java/techreborn/blockentity/storage/energy/HighVoltageSUBlockEntity.java
@@ -44,7 +44,7 @@ public class HighVoltageSUBlockEntity extends EnergyStorageBlockEntity implement
 	 * MFSU should store 4M Energy with 512 E/t I/O
 	 */
 	public HighVoltageSUBlockEntity(BlockPos pos, BlockState state) {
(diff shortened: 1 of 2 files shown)


src/main/java/techreborn/blockentity/storage/energy/HighVoltageSUBlockEntity.java
```diff
@@ -44,7 +44,7 @@
 	 * MFSU should store 4M Energy with 512 E/t I/O
 	 */
 	public HighVoltageSUBlockEntity(BlockPos pos, BlockState state) {
-		super(TRBlockEntities.HIGH_VOLTAGE_SU, pos, state, "HIGH_VOLTAGE_SU", 2, TRContent.Machine.HIGH_VOLTAGE_SU.block, RcEnergyTier.HIGH, 4_000_000);
+		super(TRBlockEntities.HIGH_VOLTAGE_SU, pos, state, "HIGH_VOLTAGE_SU", 2, TRContent.Machine.HIGH_VOLTAGE_SU.block, RcEnergyTier.HIGH, 100_000_000);
```
drcrazy: MFSU should be 40M, not 100M. Reference: https://ftb.fandom.com/wiki/MFSU_(IndustrialCraft_2)

LauchInterceptor: I think it's cheesy that the mfsu stores almost 40% less energy than the sum of its parts but yeah 100M is a bit much

Issue #2667, opened 2021-12-20
Flag #minecraft:mineable/hoe missing from techreborn:rubber_leaves


RubyRezal:
**Describe the bug**
Using a hoe doesn't break rubber tree leaves faster than other tools unlike other leaf blocks

**Steps to Reproduce**

**Screenshots**
![2021-12-19_20 42 26](https://user-images.githubusercontent.com/1816386/146700672-d74e71af-a963-464e-adde-95f7968a2f14.png)
![2021-12-19_20 42 45](https://user-images.githubusercontent.com/1816386/146700680-d6323d54-0a60-4e80-a5bc-29dcfe0ef3cf.png)

**Environment (please complete the following information with the version):**
 - Minecraft: 1.18.1
 - Mod Loader: Fabric 0.12.12
 - Tech Reborn 5.1.0-beta.4
 
**Logs**
N/A

**Additional context**
For parity with vanilla tool use mechanics.

drcrazy:
Fixed by 4e17a341c32cc7b75535eeca2edac15a1286d672

PR #2668, opened 2021-12-20
Add rubber leaves to hoe mineable blocks

LauchInterceptor:
Rubber leaves should now be mineable using hoes


LauchInterceptor:
Closes #2667

PR #2678, opened 2021-12-23
Implement partial return for damaged items

drcrazy:
This might be too hacky. Comments are welcome

### DIFF
--- a/src/main/java/techreborn/api/recipe/recipes/BlastFurnaceRecipe.java
+++ b/src/main/java/techreborn/api/recipe/recipes/BlastFurnaceRecipe.java
@@ -25,27 +25,26 @@
 package techreborn.api.recipe.recipes;
 
 import com.google.gson.JsonObject;
+import com.google.gson.JsonSyntaxException;
 import net.minecraft.block.entity.BlockEntity;
 import net.minecraft.item.ItemStack;
 import net.minecraft.util.Identifier;
 import net.minecraft.util.JsonHelper;
-import net.minecraft.util.collection.DefaultedList;
 import reborncore.common.crafting.RebornRecipe;
 import reborncore.common.crafting.RebornRecipeType;
-import reborncore.common.crafting.ingredient.RebornIngredient;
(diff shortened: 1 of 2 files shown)


LauchInterceptor:
I agree, it's a bit hacky. Problems with this are
- The recipe object is not stateless anymore ( Could cause issues? )
- When writing a recipe json you have no control over what part of the output is affected and by how much.
- It takes the first damageable item of the input. Thus when you register a recipe with more than one damageable item it can lead to unintended behaviour.

drcrazy:
Yeah, this is why it is a PR and not committed to main branch yet. 

1) State for REI cached on first load and use item with full durability. State updated right prior to crafting
2 + 3) Overall system relies on 'partialReturns' flag in recipe. Normal recipes will not have this flag, thus no changes to any normal recipe even if inputs in this recipe have durability. Armor recycling recipes in IBF should follow undocumented pattern of armor + non-damageable itemstack -> ingot(s) + some single item. This is the weakest part of implementation.

Ayutac:
maybe we can just differentiate between "damaged" and "undamaged"?

xanthian:
We had discussed this at length on discord, and reached a bit of an impasse on it.
Consensus was that current full return of items is OP.

Suggestions were around the reduction of returned items by at least 1 quarter, to save the hacky code required to implement return rates based on item dura.

Considering only 1 nugget is returned on vanilla when smelting tools/weapons, my thoughts were only to reduce the blast furnace returns considerably.

Also another thought was to add higher heat requirements for the higher tier items.

Also thought about moving Diamond items over to Grinder as its more fitting with the actual process it may go through, rather than smelting it 😁

Ayutac:
after looking at the code I think this is rather well done as a solution

But I'm still unsure if I want this in this form :(

Ayutac:
I got another idea: What about we make the return value a percentage that is independent of damage, but configurable? That way we can initialize the recipe on startup, so it remains stateless, since it is not changed during runtime?

Ayutac:
#2893 was approved and so this PR is outdated

PR #2687, opened 2021-12-27
Add tag-based recipes for plantballs and compressed plantballs

Adirelle:


Ayutac:
the plantball crafting recipe must be removed. That one is usually reserved for storage blocks. Especially, I see issues with 9 wheat -> hayball.
Rest looks good though

Adirelle:
Right now, there is no recipe for the (not-compressed) plantball. A 9-item recipe would conflict with some storage blocks. However, maybe a 8-item shapeless recipe could be ok. Using item tags allows to mix different items.

Remainder: the (not-compressed) plantball still need to be compressed in order to obtain a compressed plantball.

PR #2703, opened 2022-01-01
1.18 ae2 ore

xanthian:
update grinder & ind grinder recipes for AE2 MODID change


Technici4n:
Weren't there recipes with fluix too?

xanthian:
> Weren't there recipes with fluix too?

yup :)

PR #2712, opened 2022-01-05
Changed the number of plates back to 9

Ayutac:
maybe it was an attempt of making lapotronic stuff more expensive/balanced, but you could still get a plate from one lapis dust and changing that now out of nowhere would probably confuse/upset many players. Hence correction of this recipe for easier use


Ayutac:
Oh, wait a moment, lapis dust makes lazuli_plates not lapis_plates. What?

Ayutac:
ok yeah, different items, with the lapis_plate having no use except for chunk loader. disregard this PR!

PR #2721, opened 2022-01-06
Added Steel Armor. Textures by Xanthian.

Ayutac:
Steel's durability is about between iron and diamond, it is almost as tough as diamond, half as tough as netherite, but has same knockback resistance as netherite (which diamond lacks completely). But because we mixed some dirty coal to our iron to make it this tough and resistant, the enchantability is down to 5, which is by far the worst enchantability of all vanilla and TR armors.

Not tested yet, but I basically did the same changes as in the silver armor commit line by line.


xanthian:
I APPROVE!!!!
(and i tested it)

PR #2729, opened 2022-01-08
Added calcite related recipes.

Ayutac:
Successfully tested.

### DIFF
--- a/src/main/resources/data/c/tags/items/coral_blocks.json
+++ b/src/main/resources/data/c/tags/items/coral_blocks.json
@@ -1,0 +1,7 @@
+{
+	"replace": false,
+	"values": [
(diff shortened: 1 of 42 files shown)


xanthian (commented):
cant say im a fan of the dripstone / calcite stuff, doesnt really make sense in MC terms they are totally different blocks.
irl/science terms I guess you're going down the calcium (calcite being calcium carbonite) route for what forms stalactite / stalagmite ?

Ayutac:
> sense in MC terms, but irl scientifically

Uhm, TR is allowed to borrow from science, right? I see no need to make "dripstone dust" that effectively behaves exactly as calcite dust does when centrifuged because they are effectively the same. Different blocks are allowed to grind into the same dusts, yes?

PR #2732, opened 2022-01-08
Misc recipe additions/changes

Ayutac:
Added smooth basalt to basalts that can be centrifuged.
Added centrifuge recipes for rooted dirt and coarse dirt.
Added centrifuge recipe for moss block, which gives (uncompressed) plant ball.
Changed centrifuge recipe for grass block to give uncompr. plant ball instead of wheat seeds, eliminating the implication that every grass block holds wheat seeds.
Added grinder recipe to get small glowstone dust out of shroomlights (I forgot that in the fungus update)
Changed grinder recipe for gravel, now accepts all stone tool materials instead of only cobblestone.
Added a use for the andesite/diorite/granite dusts. Craft all 3 together with flint to get 4 gravel.
Added crafting 9 leaves/saplings of any kind into a plantball.
Added extractor recipe for green dye by using uncompr. plantball. Should never allow compressed plantball as that is out of juice, so to say.
Added extractor recipe for orange dye by using carrots.
Added extractor recipe for red dye by using beetroot.
Added extractor recipe for hanging roots by using rooted dirt.
Changed extractor recipe for stick to yield two sticks instead of one.
Changed electrolyzer recipe to accept anything tagged as sand instead of only the item sand, now also allowing red sand to be electrolyzed into silicon.

### DIFF
--- a/src/main/resources/data/techreborn/recipes/centrifuge/coarse_dirt.json
+++ b/src/main/resources/data/techreborn/recipes/centrifuge/coarse_dirt.json
@@ -1,0 +1,25 @@
+{
+  "type": "techreborn:centrifuge",
+  "power": 5,
+  "time": 2500,
+  "ingredients": [
+    {
+      "item": "minecraft:coarse_dirt",
+      "count": 16
+    }
+  ],
+  "results": [
+	{
+	  "item": "minecraft:sand",
+	  "count": 2
+	},
+	{
+	  "item": "minecraft:dirt",
+	  "count": 5
+	},
+    {
+      "item": "minecraft:gravel",
+      "count": 9
+    }
+  ]
+}
--- a/src/main/resources/data/techreborn/recipes/centrifuge/coarse_dirt.json
+++ b/src/main/resources/data/techreborn/recipes/centrifuge/coarse_dirt.json
@@ -10,16 +10,15 @@
   ],
   "results": [
 	{
-	  "item": "minecraft:sand",
-	  "count": 2
+	  "item": "minecraft:gravel",
+	  "count": 8
 	},
 	{
-	  "item": "minecraft:dirt",
-	  "count": 5
+	  "item": "minecraft:flint"
 	},
     {
-      "item": "minecraft:gravel",
-      "count": 9
+      "item": "minecraft:sand",
+      "count": 2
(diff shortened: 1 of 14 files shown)


modmuss50 (approved):
looks fine at a glace, will wait for someone else to take a look at the recipe changes balance wise and what not 👍

xanthian (commented):
the sapling > plantball should return more than leaves > plantball
otherwise looks good

Ayutac:
> the sapling > plantball should return more than leaves > plantball otherwise looks good

Good point. 2 or 3? I'm in favour of 3.

xanthian:
more - 10?
its only 10% chance to get sapling from leaves with fortune 3

src/main/resources/data/techreborn/recipes/centrifuge/coarse_dirt.json
```diff
@@ -0,0 +1,25 @@
...
+  "type": "techreborn:centrifuge",
+  "power": 5,
+  "time": 2500,
+  "ingredients": [
+    {
+      "item": "minecraft:coarse_dirt",
+      "count": 16
+    }
+  ],
+  "results": [
```
drcrazy, line 11: This could be 8xSand + 2xGravel + Flint to be in line with dirt

Ayutac:
> more 

You say 10% chance for sapling per leave, I say 30 new leaves with one sapling. But if you insist... 5?

src/main/resources/data/techreborn/recipes/centrifuge/coarse_dirt.json
Ayutac: didn't think of that one. my idea was that coarse dirt has a lot more gravel in it and there is no centrifuge recipe for gravel. Maybe 8 Gravel, 2 Sand, 1 Flint?
Anyway, I do not have strong feelings about this recipe, do as you like

drcrazy: Yeah, 8xGravel, 2xSand and Flint looks good. Go for it.

Ayutac:
Forgot the podzol lol. Also changed the other two debated recipes

PR #2733, opened 2022-01-08
Make energy tools usable while charging, fixes #2232

### LINKED ISSUE #2232: Advanced jackhammer starts to work faster when the lithium batpack is discharged
**Steps to Reproduce**
Watch the speed of the jackhammer. Break blocks with an advanced jackhammer until the lithium battery pack is discharged. You will see the block breaking speed increase.

**Screenshots**
Imgur video: [tap](https://imgur.com/a/jmhiksw)

* Minecraft version: 1.16.3
* Mod Loader: fabric-loader-0.9.3+build.207-1.16.3, fabric-api-0.21.0+build.407-1.16, RebornCore-1.16-4.4.2+build.73, TechReborn-1.16-3.5.4+build.122

### LINKED ISSUE #2370: When charging mining tool, it is unuseble
**Describe the bug**
When digging tool is charging, this is nearly unusuble because he changes.

**Steps to Reproduce**
- Any enabled battery or batpack in inventory
- Any diging tool from TechReborn
- Try to dig when tool blinks while NTB changes (charging)

 - Minecraft: 1.16.5
 - Mod Loader: Fabric

**Additional context**
I think the problem is in NTB and Minecraft behaviour when it changes.
But IC2 and other Forge mods fixed it.

vladislemon:
Used Fabric API features to prevent constant block breaking canceling while ItemStack in hand charges.
Fix for #2232 and closed #2370 

Before: https://youtu.be/TUTQsJ36TdQ
After: https://youtu.be/ej2ADwSasrk

### DIFF
--- a/RebornCore/src/main/java/reborncore/common/powerSystem/RcEnergyItem.java
+++ b/RebornCore/src/main/java/reborncore/common/powerSystem/RcEnergyItem.java
@@ -1,5 +1,10 @@
 package reborncore.common.powerSystem;
 
+import net.fabricmc.fabric.api.item.v1.FabricItem;
+import net.minecraft.entity.player.PlayerEntity;
+import net.minecraft.item.ItemStack;
+import net.minecraft.util.Hand;
+import reborncore.common.util.ItemUtils;
 import team.reborn.energy.api.base.SimpleBatteryItem;
 
 /**
@@ -10,7 +15,7 @@ import team.reborn.energy.api.base.SimpleBatteryItem;
  * </ul>
  * TODO: consider moving this functionality to the energy API?
  */
-public interface RcEnergyItem extends SimpleBatteryItem {
+public interface RcEnergyItem extends SimpleBatteryItem, FabricItem {
 	long getEnergyCapacity();
 
 	/**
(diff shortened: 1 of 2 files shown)


src/main/resources/fabric.mod.json
```diff
@@ -39,6 +39,9 @@
 	"team_reborn_energy": ">=2.0.0-beta1",
 	"fabric-biome-api-v1": ">=3.0.0"
   },
+  "mixins": [
+	"techreborn.mixins.json"
```
modmuss50: Ideally these mixins should go into RebornCore (Just a sub project), we try to keep TR free of mixins. Should hopefully not be too hard to do this, if you are unsure and need some pointers please let me know. 👍 

Otherwise it looks resonable at a glance.

modmuss50 (changes requested):

src/main/resources/fabric.mod.json
vladislemon: Thanks!
I moved changes to RebornCore (and formatted mixins files a bit for correct automatic line sorting)

modmuss50:
Just having a quick look, I think Fabric already provides APIs to do this? See: https://github.com/FabricMC/fabric/blob/ded849a99973c60b24d1471a645f5ebdb803b3d8/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItem.java

Or are these not the same?

vladislemon:
> Just having a quick look, I think Fabric already provides APIs to do this? See: https://github.com/FabricMC/fabric/blob/ded849a99973c60b24d1471a645f5ebdb803b3d8/fabric-item-api-v1/src/main/java/net/fabricmc/fabric/api/item/v1/FabricItem.java
> 
> Or are these not the same?

You're right, it's same thing.
I'm not very familiar with Fabric API, missed that.
Thank you very much! This makes code a lot cleaner.

modmuss50 (approved):
Looks great now 👍

PR #2734, opened 2022-01-09
Attempt to fix #2625

### LINKED ISSUE #2625: No icon on the Scrapbox REI tab
No icon on the Scrapbox REI tab

Fabric 1.18-rc4
TR 5.1.0 b2

same thing on earlier 1.18 versions of both ^^

Ayutac:
For details see the discussion of that bug.


Ayutac:
Note, this is NOT yet fixed, do NOT merge into project. This PR is to reference in the bug discussion. Maybe we can mend it to a proper fix, but for now pls just let it be and don't close either.

modmuss50:
Personally I would just change the lang file, and leave the code change.

Ayutac:
Fixed instead by 7e1125f

PR #2737, opened 2022-01-09
Cell dupe fix

### LINKED ISSUE #2664: Chemical Reactor - Cell Dupe
The following recipes return more cells than you put in, effectively duping them.
Rather pointless, but surely an oversight.

image
image

Issue is in both 1.17.1 & 1.18.1.

xanthian:
amended recipes to return correct number of cells as per input.
reduced 3, increased 2.
fixes #2664


xanthian:
Could I get some feedback on this please :) ?

Happy to make further changes such as reducing craft time of those that have been effectively 'nerfed'

modmuss50 (approved):

PR #2739, opened 2022-01-10
Recipe tweaks

xanthian:
got more in the pipeline, but will commit these for now.
thoughts please..
my only thought is with gunpowder, is 64 too high? (its easy to farm tbh..)


Ayutac:
Is the chorus fruit one supposed to be a centrifuge recipe? And independent of that, why produce nitrogen?

The chemical reactor recipe with magma cube makes no sense, since there is nothing reacting. Or anything, since nbt isn't specified. Also why would the result be lithium?

Gunpowder has a shapeless recipe and that doesn't match the centrifuge one. I don't see how the centrifuge would disconnect the KNO3. TR uses an electrolyzer for exactly that job, see the recipe for saltpeter_dust there.

Take 10 carbon & 10 calcium, chemical reactor to 20 calcium carbonate, your recipe to 20 calcite_dust and 20 carbon. 20 calcite_dust get electrolyzed to 4 calcium, 4 carbon and 6 compressed air. 14 carbon out of 6 calcium sounds broken. Which is because Calcium carbonate and calcite are basically the same thing.
I could imagine increasing the calcite_dust in the recipe a bit, say to 3 or 4, but adding carbon seems way too much.

The two prismarine grinder recipes are okay.

Ayutac:
Calcium carbonate is okay now. 

Also I have finally read the full commit logs.

I would still remove nitrogen from the chorus fruit recipe, especially since it returns ender dust which can be processed into nitrogen. I like the reasoning behind the endstone dust, but if it is just a trace element, I would return 2 endstone_small_dust instead, similarily as I did with glow berries and glowstone_small_dust

I do not find a good source for the lithium/magma claim. Instead this: https://en.wikipedia.org/wiki/Lithium#Terrestrial

So I would rather suggest an electrolyzer recipe of granite dust producing besides other things lithium. But furthermore, I would let lithium as it is. Clay itself is not everywhere, but it can be centrifuged out of dirt, so while lithium might only have "one" source, it is the most abundant one after stone variants...

xanthian:
needs a rethink.

Ayutac:
feel free to DM me about ideas

Issue #2617, opened 2021-11-27
Some Tooltips too large for GUI scale 3 & 4 (1.18 pre4)


xanthian:
Example
![image](https://user-images.githubusercontent.com/7688001/143664297-f8d4a5bf-3487-4414-bc36-65599ac04607.png)

Screen Resolution 1920x1080

List of machines with issue at scale 3 (when holding shift)

Vacuum Freezer
Lightning Rod
Drain
Windmill
Industrial Chunkloader

List of machines with issue at scale 4

All above +
Extractor
Implosion Compressor
Industrial Electrolyzer
Industrial Grinder
Iron Alloy Furnace
Solid Canning Machine
Recycler
Gas Turbine
Plasma Generator
Semi-Fluid Generator
Generator
Player Detector
All Iridium Reinforced Tungsteel items

worth noting that in some cases depending on the toolbar or bag slot the item is placed, it may display most if not all, but this is not ideal.

drcrazy:
Vanilla issue. There are a number of mods like https://www.curseforge.com/minecraft/mc-mods/tooltipfix to fix it.

PR #2741, opened 2022-01-12
Update lang file

xanthian:
Numerous changes, 
Mostly to resolve the large GUI issues reported in #2617 
Re-ordered some tooltips to read from Top to bottom rather than bottom to top.
Shortened and reworded some so they are more to the point.
Correct spelling mistakes.
Entries that were relevant but more of a minor not were made grey and italic, this can be removed if rejected.


modmuss50:
Seems like you may have changed the line endings, hiding the whitespace changes makes it easier to review.

PR #2751, opened 2022-01-14
Added sponge piece + recipes, texture by Spearkiller

Ayutac:
Okay, hear me out.

Elder guardians drop sponge and cod.
Guardians drop cod.

Hence there is something in cod that makes sponge, right?? Minecraft-logic!

Balancing: If you got a guardian farm, you get infinite cod and hence infinite, but the cod rates are rather low. If you afk fish, you will get a looooot of cod faster. But for normal players, getting a lot of cod (9 stacks for one sponge) it takes a lot of grinding time. So this is an alternative way to get sponge if a player doesn't want to go into an ocean monument, but it is very grindy. On the other hand, if you fish a lot normally (not afk), eg to get treasure loot, you get a loooooot of cod that isn't all too useful. Getting sponge for that seems like a nice extra.

You could reduce the amount to 48 or 32 of course, but imo that would mainly increase industrial production.


xanthian:
Cod is equally as common drop as prismarine crystals / shards from Guardians
image
This is no more than a couple of hours afk.

I partly disagree with this recipe.
The sponge part is a decent start, but cod is just too easy / and a bit random

Ayutac:
hm, I thought it was less. What about salmon or tropical fish then? I would like a single source to farm. the sulfide recipe can come later.

where can one extract sponge from?

xanthian:
> hm, I thought it was less. What about salmon or tropical fish then? I would like a single source to farm. the sulfide recipe can come later.
> 
> where can one extract sponge from?

salmon is a fresh water fish, so wouldn't make any sense.
tropical.. possibly? much harder to obtain en-masse (without a fish farm.. but who builds those) still seems weird tbh.

modmuss50 (approved):

PR #2758, opened 2022-01-18
Added optional removal of the current recipe unlock

Ayutac:
Should default to false in old and new configs, ie no new behaviour, but now I can deactivate it via config instead of using an unofficial distribution. Will probably come back to this once the toasts are finished...


Ayutac:
oh, and if you could release a beta after this PR was accepted, that would be great :)

modmuss50 (approved):

Issue #2763, opened 2022-01-20
Data generated with datagen not used in dev


Ayutac:
The data generated with datagen is not used in dev environment. This makes checking for correct behavior way more tedious.

Ayutac:
I guess that is also the reason why https://github.com/Ayutac/TechReborn/tree/1.18-Ayutac-datagen didn't work. The generated tags are not added to TR in dev env and hence the static references to them in the JSON files cause errors. Hrmmm

modmuss50:
Humm, odd ill take a look.

Ayutac:
Fixed by 646132e9 (and following commits)

PR #2764, opened 2022-01-20
Added dust crafting data gen

Ayutac:
Datagen works, even if not testable within dev env, see #2763.

Manual old dust <-> small dust recipe files have been deleted, as they are now provided by datagen. Additionally, their toasts - which did previously not exist - are generated as well.


Ayutac:
With datagen changes this doesn't work anymore

PR #2769, opened 2022-01-23
1.18 ayutac datagen4

Ayutac:
datagen, wohoo!


src/datagen/groovy/techreborn/datagen/recipes/TechRebornRecipesProvider.groovy
```diff
@@ -50,13 +50,16 @@ abstract class TechRebornRecipesProvider extends FabricRecipesProvider {
     abstract void generateRecipes()
 
     static Ingredient createIngredient(def input) {
-        if (input instanceof ItemConvertible) {
+        if (input instanceof Ingredient) {
+			return input
+		}
+		if (input instanceof ItemConvertible) {
```
modmuss50, line 56: Add TagConvertible before ItemConvertible? Should let you remove all of the `.asTag()` calls unless im missing something?

modmuss50:
This exception seems to have been introduced when running the test server? <https://github.com/TechReborn/TechReborn/runs/4912125016?check_suite_focus=true#step:5:269>

Not quite sure why.

src/datagen/groovy/techreborn/datagen/recipes/TechRebornRecipesProvider.groovy
Ayutac: I'm unsure if we want that. There might be instances where it is better to call the TR id instead of the common tag. So using the tag as of now requires a conscious decision

Ayutac:
looks like techreborn/tags/ingots.json wasn't generated?

src/datagen/groovy/techreborn/datagen/recipes/TechRebornRecipesProvider.groovy
modmuss50: Ok, thats fine 👍 can always change it later.

modmuss50:
Its in the jar, I expect its an issue with the gametest module not having the data gen resources. Im happy to ingore that for now. 👍

modmuss50 (approved):

Ayutac:
it really *should* work. I would look at it, but IntelliJ has suddenly decided not to recognize the existence of TRContent anymore.

Ayutac:
issue was that client dev test doesn't run datagen.

Working of the generated files has been affirmed by Xanthian

PR #2773, opened 2022-01-23
Quick hacky fix for Modded overworld biome compatibility (terralith+@)

aria1th:
it does NOT expect custom dimensions being added


aria1th:
No, just wait for better fabric API...

PR #2777, opened 2022-01-30
Resin Basin updated. Thanks to Spearkiller and Ayutac.

Ayutac:
Hitbox has been updated to half a block, better matching the model.
Item model has been updated by Spearkiller to look nice.
There was a weird inconsistency about the amount of sap given to player depending on either putting it into an inventory or breaking the block.

Right click to harvest basin didn't work as intended, so removed for now.

Also it seems the block states (full/pouring/pouringTime) are not saved via nbt?

Also also wow the warnings.

### DIFF
--- a/src/main/java/techreborn/blocks/machine/tier1/ResinBasinBlock.java
+++ b/src/main/java/techreborn/blocks/machine/tier1/ResinBasinBlock.java
@@ -34,6 +34,9 @@ import net.minecraft.state.property.BooleanProperty;
 import net.minecraft.state.property.DirectionProperty;
 import net.minecraft.state.property.Properties;
 import net.minecraft.text.TranslatableText;
+import net.minecraft.util.ActionResult;
+import net.minecraft.util.Hand;
+import net.minecraft.util.hit.BlockHitResult;
 import net.minecraft.util.math.BlockPos;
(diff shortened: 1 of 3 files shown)


src/main/java/techreborn/blocks/machine/tier1/ResinBasinBlock.java
```diff
@@ -82,6 +88,19 @@ public Direction getFacing(BlockState state) {
 		return state.get(FACING);
 	}
 
+	/*@SuppressWarnings("deprecation")
+	@Override
+	public ActionResult onUse(BlockState state, World world, BlockPos pos, PlayerEntity player, Hand hand, BlockHitResult hit) {
```
modmuss50, line 93: Is this needed? Generally dont want to commit commented out code?

Ayutac: The idea is that it *should* work, but it doesn't. I don't want to commit a bug, but I want people to be able to solve the bug, because I am unlikely to follow it further.

Ayutac:
I fixed the out-commented bug by myself now :)

modmuss50 (approved):

Issue #2621, opened 2021-11-28
Chrome should be renamed to chromium (the correct English word)


Technici4n:
We'll rename it in MI for 1.18.

drcrazy:
What chrome items should be chromium? It looks like 'Chrome stairs' is fine, while 'Block of Chrome' probably should be 'Block of Chromium'.

xanthian:
I think it would cause more issues renaming it now, maybe for next patch 1.19?

But..
Nugget, Ingot, Small Dust, Dust, Block are the obvious ones.
Chrome Plate makes sense to leave as is although 'Chromium Plating' is an uncommonly used name for it.

Chrome Plated block, stairs, slab and wall would initially make sense, but as they are made with Chromium, they are not Chromed  
So all items really 🤣

Chrome is actually the name of product created by plating a metal with Chromium.

A few addendums..

Chromite (raw form of Chromium) should really be added as an Ore, as its technically not a by-product of anything
Processing of Chromite would produce Chromium and Iron.
Why it is pink

Ayutac:
Renaming would happen in the lang file, there would be no other issued I could foresee rn

xanthian:
that would make sense lol
other points stand.

xanthian:
draft PR in for this if its wanted.
#2780

Ayutac:
Solved by #2795

PR #2780, opened 2022-01-31
Rename Chrome to Chromium

xanthian:
rename chrome > chromium #2621


xanthian:
unsure if this was wanted/needed.
feel free to reject if not.

modmuss50 (approved):
Seems fine for me

Technici4n:
Which tag did we use in MI again? Might be `c:chromium_blocks` or so.

xanthian:
> Which tag did we use in MI again? Might be `c:chromium_blocks` or so.

good point, not added/updated tags, dont merge this yet.

Ayutac:
NO! We do NOT change established tag names. I don't care if the displayed name is changed, but leave your fingers off variable names.

We don't need to create ambiguity with tags. This will only lead to different mods using different variants. Keep it clean, stick to lang file only.

Technici4n:
MI is already using `c:chromium_xxx` tags though. Changing the tag in TR will reduce the ambiguity...

Ayutac:
@Technici4n can you check the other main tech mods for their tag? might do it right for all at once, if we do it at all

Technici4n:
I believe only MI and TR have chromium materials.

PR #2791, opened 2022-02-07
Fixes #2783

### LINKED ISSUE #2783: Raw variants take 10 seconds in vanilla blast furnace
I dont know if this is intended behavior, and if it is so, then this isn't an issue. but every recipe that uses raw variants of ores inside the vanilla blast furnace takes 10 seconds, while consuming twice the amount of fuel necessary. eg, coal smelts 4 items instead of 8. I should clarify that this only affects the mod variants and not the vanilla raw ores themselves. It also seems to override other mod recipes of the same ore and causes this issue with other mod variants, for example, raw tin from another mod takes 10 seconds as well. Removing tech reborn reverts to the other mods recipe which only takes 5 seconds.

Ayutac:
Needs to be tested still. Goal: changed blasting to use half as much as time as regular furnace for same item.

EDIT: Tested by ErrorSys and debugged what they found. Can be merged now.


ErrorSys:
Compiles and works as intended, although I did have to specify experience on my end before IDEA would compile. Probably just on my end though as it should be an implied integer anyways.

PR #2798, opened 2022-02-09
Added 2 canning machine recipes for coolant cells

Ayutac:
Hopefully this can ease the construction of those.


drcrazy:
Solid canning machine has two inputs. Anyway, I will fix it.

Ayutac:
I was under the impression that the second input was optional. Goal was to circumvent the use of a bucket without changing the materials that would be needed. But I guess this works too :)

PR #2799, opened 2022-02-10
1.16 fix backport

Starjumper2821:
Don't actually want a merge. But if someone could tell me if I'm being more stupid than usual before I continue altering the code that would be splendid.


modmuss50:
Im not too sure what you are trying to do. We dont support 1.16 anymore.

PR #2818, opened 2022-02-15
Added hardness and blast resistance to metal blocks

### LINKED ISSUE #2786: Iridium-Reinforced Tungstensteel Block Blast Resistance
Fabric Loader 0.12.12-1.18.1
fabric-api-0.46.1+1.18
TechReborn-5.1.0-beta.8
RebornCore-5.1.0-beta.8

Wondering if the IRTB blocks have had their blast resistance dropped/altered.

Created a TNT block encased in a square of IRTB, and it shattered the IRTB back to floating entities.  Did a secondary test of it with vanilla Obsidian surrounding TNT block, and the Obsidian was untouched.

ErrorSys:
Should address and fix #2786, needs to be tested though.

Created hardness and resistance values for metal blocks respective of their types.
Also added properties for individual machine casings so that they are easier to break with a pickaxe.


Ayutac:
Does that mean "techreborn:basic_machine_casing" and the other casings don't drop when broken without pickaxe anymore?

PR #2824, opened 2022-02-21
Port to 1.18.2

modmuss50:
Should be fully featured, didnt have any issues with missing API from Fabric.


Ayutac:
*looks at failed checks*
should we be worried?

PR #2832, opened 2022-02-26
Big datagen/toast update!

Ayutac:
So this is it, the reason I joined this project ;-)

Proper toasts and, to my own surprise, more datagen. It's beautiful.

Might contain a few bugs still. I remember in my test run I got some free recipes for no good reason. But this can be fixed with time. What's important to me is that we merge this because TRContent gets frequently updated and I have to work around that every time. Small bug fixes should usually do without changing TRContent.


modmuss50:
I will merge 1.18.2 into this, and review at the same time. Looks awesome at a glace. Might take me a day or two to get around to doing it. 👍

Ayutac:
Well, it was your ideas and stubs that made it possible for me, so you got to praise yourself a bit as well 😌

If there are any questions, ask away

modmuss50:
Merge was easy enough, I did have to make a small fix to TrContent.Plates to make it compile.

PR #2833, opened 2022-02-26
Prevent Frost Walker freezing Mod Liquids

aria1th:
Side effect : it gets viscosity instead but no crash happens.


Ayutac:
Have you tried walking on it with a strider? :D 

This is a hacky solution. The problem lies within FrostWalkerEnchantment, because the blockstate is checked to be of material water. I would prefer solving this rigorously. Eg. electrolyzed water could still become frosted ice, helium plasma could become helium, hot tungstensteel could become normal tungstensteel. Would be fun.

best would probably be to introduce TR's own liquid materials

drcrazy:
+1 to Ayutac. I would rather prefer custom freeze logic.

PR #2840, opened 2022-03-01
add c:workbench to autocrafter

xanthian:
to include compat for byg, betterx and other mods with added crafting tables
tested working in dev, vanilla crafting table unaffected by change

### DIFF
--- a/src/main/resources/data/c/tags/items/workbench
+++ b/src/main/resources/data/c/tags/items/workbench
@@ -1,0 +1,4 @@
+{
+  "replace": false,
+  "values": ["minecraft:crafting_table"]
+}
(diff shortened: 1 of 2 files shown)


Ayutac (changes requested):
I would recommend only leaving the c:workbench in the recipe and add the tag to our common tags. Is cleaner.

xanthian:
what changes are required here? 
no reason not to leave both there, alleviates any potential issues with tags

xanthian:
added new c tag for workbench to include vanilla table, removed entry from recipe to make it 'cleaner'

Ayutac (approved):

PR #2841, opened 2022-03-01
Allow Drain drain Cauldrons

aria1th:
There's no proper fluid interface in cauldron block so its only working for Water and Lava. If any mods uses cauldron as other fluid holder, it might have to change, but I've never seen one.

### DIFF
--- a/src/main/java/techreborn/blockentity/machine/misc/DrainBlockEntity.java
+++ b/src/main/java/techreborn/blockentity/machine/misc/DrainBlockEntity.java
@@ -24,11 +24,10 @@
 
 package techreborn.blockentity.machine.misc;
 
-import net.minecraft.block.Block;
-import net.minecraft.block.BlockState;
-import net.minecraft.block.FluidDrainable;
+import net.minecraft.block.*;
 import net.minecraft.entity.player.PlayerEntity;
 import net.minecraft.fluid.Fluid;
+import net.minecraft.fluid.Fluids;
 import net.minecraft.item.ItemStack;
 import net.minecraft.util.math.BlockPos;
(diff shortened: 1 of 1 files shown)


Technici4n:
Just use a tank unit and import from the top?

src/main/java/techreborn/blockentity/machine/misc/DrainBlockEntity.java
```diff
@@ -89,6 +88,26 @@ private void tryDrain() {
 				TechReborn.LOGGER.debug("Could not get Fluid from ItemStack " + fluidContainer.getItem());
 			}
 		}
+		if (aboveBlock instanceof LeveledCauldronBlock && aboveBlockState.isOf(Blocks.WATER_CAULDRON)) { //ensure Water cauldron
```
Ayutac, line 91: what is a LeveledCauldronBlock and why don't you need that check for the lava?

Ayutac:
> Just use a tank unit and import from the top?

@Technici4n can a tank unit access cauldrons?

Ayutac (approved):
Code looks valid, just unsure if this feature should be added, given Technici4n's comment.

fooooooooooooooo:
Having two ways of extracting fluids from cauldrons isn't a bad thing, and it makes logical sense that the drain should be able to drain a cauldron. In my opinion this is a good feature.

Technici4n:
I don't like the way this works though, at this point why check for cauldrons specifically... This could use the fabric fluid API to pull from any fluid container above.

drcrazy:
Let's start with cauldrons only and them switch to FAPI

PR #2843, opened 2022-03-02
Cache BFS Result and use when its possible

aria1th:
Reduces 8.5mspt -> 5.5mspt for 9300 glass fiber cables.

Confirmed working in sp / mp but test / review required, there might be edge case I'm not aware of.

### DIFF
--- a/RebornCore/src/main/java/reborncore/common/blockentity/SlotConfiguration.java
+++ b/RebornCore/src/main/java/reborncore/common/blockentity/SlotConfiguration.java
@@ -25,13 +25,18 @@
 package reborncore.common.blockentity;
 
 import com.mojang.brigadier.exceptions.CommandSyntaxException;
+import net.fabricmc.fabric.api.lookup.v1.block.BlockApiCache;
 import net.fabricmc.fabric.api.transfer.v1.item.InventoryStorage;
 import net.fabricmc.fabric.api.transfer.v1.item.ItemStorage;
+import net.fabricmc.fabric.api.transfer.v1.item.ItemVariant;
+import net.fabricmc.fabric.api.transfer.v1.storage.Storage;
 import net.fabricmc.fabric.api.transfer.v1.storage.StorageUtil;
 import net.minecraft.inventory.Inventory;
 import net.minecraft.item.ItemStack;
 import net.minecraft.nbt.NbtCompound;
 import net.minecraft.nbt.StringNbtReader;
+import net.minecraft.server.world.ServerWorld;
+import net.minecraft.util.math.BlockPos;
 import net.minecraft.util.math.Direction;
 import org.apache.commons.lang3.Validate;
 import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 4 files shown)


aria1th:
I guess it might require blockEntity.isRemoved() check for keys and remove it too.

Technici4n (changes requested):
It looks like this is going to leak a large amount of memory...

I am really not convinced by this optimization, halving the tick time is not that impressive for something that should be a large complexity improvement?

I would suggest profiling the algorithm to check if something can be made faster inside of the BFS (getBlockEntity can be made much faster with BlockApiCache).

aria1th:
Profiles with spark : before opt
https://spark.lucko.me/CgPVQDSCAh
After opt
https://spark.lucko.me/5SRQtHpzf6
and yes I agree if getBlockEntity can be speed up, it will be much faster... 

And I checked if *just* BlockApiCache can help it, seems its answer is no, well its almost costing 42mspt ... or I misimplemented it? actually without any patch it costs 53mspt yes....
https://spark.lucko.me/LgfBD6ARgI
https://spark.lucko.me/LDciduPLNM

but I'll add ApiCache because its very fast

aria1th:
and yeah its vanilla bug makes it 40mspt... 
Comparison without Autosaving : 
HashMap cache based : https://spark.lucko.me/m5UFU5jT8c
No cache based : https://spark.lucko.me/K2s835uB5Z

Technici4n:
Well you're not using BlockApi**Cache** as far as I can tell?

Technici4n:
It should be fine to just re-use whatever cache is used for the energy API already.

aria1th:
it wasn't blockentity cache (or I was misleading something?) so I implemented with hashmap cache and works pretty well,
https://spark.lucko.me/uHt2AWSE5h

So there's totally no getBlockEntity lag now, but most of the lag is coming from simulation of transfer amount and sorting, when cable has lower energy than required. Do we really need to prioritize cable transfer? I mean, real electricity never works that way. By removing sort / shuffle and just implementing simple iteration, it can totally remove cable lag

aria1th:
https://spark.lucko.me/bMvVvxqo0I
It was missing cache for blockentity Itself, added it and now its 10mspt with 9300 cables

There's no memory leak : 
Memory report :
https://spark.lucko.me/fZ3q9PMuXQ
https://spark.lucko.me/9AhGfD8Hd1 (40min with intensive fill repeat for 10k cables)

Technici4n:
> Profiles with spark : before opt
https://spark.lucko.me/CgPVQDSCAh
After opt
https://spark.lucko.me/5SRQtHpzf6
and yes I agree if getBlockEntity can be speed up, it will be much faster...

First profile, relevant part:
image

Second profile, relevant part:
image

It seems that there is actually no optimization?

Technici4n:
> Do we really need to prioritize cable transfer? I mean, real electricity never works that way. By removing sort / shuffle and just implementing simple iteration, it can totally remove cable lag

Yes, cables properly splitting energy is a feature that should be kept.

aria1th:
its 6:4 to 7:3 optimization so minor, and that didn't contain proper caching.. now it does it properly. BFS caching only reduces lag by 33%, but yes as you said, Block Entity was major problem and caching it solved everything. Thank you for BlockApiCache, its very promising. I'll update final report in this comment.
https://spark.lucko.me/7Y8Dd3U1yq

Technici4n:
#2844 is how I expected it to be done.

I also don't see how your PR optimizes the BFS since the cache is cleared every tick.

aria1th:
No, its not cleared every tick - it only calls tickCache if gatherCable (BFS search) is required. Look at https://spark.lucko.me/bMvVvxqo0I
gatherCable method is never called when BlockEntity is not dirty.

Technici4n:
Ah, my bad. I am still very confused by the `HashSet<HashSet<>>` and I find your code quite hard to understand. I also have concerns about memory being properly cleared when the whole cable set is removed in one tick. And finally you don't seem to properly check if cables are in a ticking chunk or not, which you should to avoid chunk loading loops.

Overall I think this PR is unnecessary given #2844.

aria1th:
Well I tested extreme cases by filling 25x25x25 cables by command block and confirmed it didn't leak any memory too, 

> https://spark.lucko.me/bMvVvxqo0I It was missing cache for blockentity Itself, added it and now its 10mspt with 9300 cables
> 
> There's no memory leak : Memory report : https://spark.lucko.me/fZ3q9PMuXQ https://spark.lucko.me/9AhGfD8Hd1 (40min with intensive fill repeat for 10k cables)

It consumed 56MB constantly, and memory didn't leak, I ran test for 5 hours and no mspt / ram usage increase was found.
But I agree its hard to read and yeah chunkloading check was missing... I hope your pr being merged quickly, since cables are really laggy in current impl.

PR #2844, opened 2022-03-03
Cable optimization: use BlockApiCache to get adjacent BE

Technici4n:
The optimization relies on `BlockApiCache#getBlockEntity` caching the target block entity, which completely eliminates the BE queries from the world starting from the second tick.


aria1th:
Can you add same method to SlotConfiguration? it should be cached too

Technici4n:
The spark profiles you linked suggest that the main issue with SlotConfiguration is the actual transfer and not the storage query, and the transfer should be sped up quite a bit once FabricMC/fabric#1979 will be merged.

PR #2845, opened 2022-03-04
Cache Torus.generate result for Origin

aria1th:
Fusion Coils are laggy, not by ticking block entities, just purely by torus.generate(BlockPos.origin, size)

Int2ObjectMap can be used alternatively tho.

### DIFF
--- a/RebornCore/src/main/java/reborncore/common/util/Torus.java
+++ b/RebornCore/src/main/java/reborncore/common/util/Torus.java
@@ -24,11 +24,13 @@
 
 package reborncore.common.util;
 
+import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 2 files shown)


modmuss50:
I would do this in genSizeMap similar to how the size is calculated. And then have a `List<BlockPos> getOriginPositions(int radius)` or something.

aria1th:
sounds good, so we can have cache at start and not generate at runtime. Can I just put calculation in genSizeMap and put in Int2ObjectMap? It uses exactly same methods, so we can generate them together.

modmuss50 (approved):
Much better 👍 Thanks a lot.

PR #2849, opened 2022-03-07
Fix RecipeCrafter spamming sound every tick

aria1th:
Recycler was able to spam sound due to this. Limit this


Ayutac (changes requested):
Couple of things:
1. make sure to not change intend style anywhere
2. don't write things you think are TODOs in a PR that has nothing to do with it
3. **WTF did you change the speedmultiplier in this PR?**

As it is now, this PR will not be accepted. The idea per se is okay. Although I haven't tested it yet myself

Ayutac (commented):
Much better :)

I would still like some dev to test it, but now it's clean.

Ayutac (approved):

PR #2866, opened 2022-03-25
Fix #2850 Fetching cooking time based on recipe

### LINKED ISSUE #2850: Iron Furnace smelts raw ore blocks too fast
**Describe the bug**
With default settings, the Iron Furnace smelts raw ore blocks in 8 seconds, but it should take 60 seconds.

**Steps to Reproduce**
1. Place an Iron Furnace.
2. Put a Block of Raw Iron in its top slot, and an Oak Slab in its bottom slot.
3. Observe that the single slab is more than enough to smelt the entire block.

 - Minecraft 1.18.1
 - Fabric Loader 0.13.3
 - Tech Reborn 5.1.0

**Additional context**
I think the issue is in AbstractIronMachineBlockEntity on line 58, where it uses a hardcoded 200 instead of checking the recipe time.

SimonFlapse:
These changes affect AbstractIronMachineBlockEntity and all subclasses. A new abstract method has been added as replacement for the `totalCookingTime` field, to facilitate the subclasses to provide the cooking time, either by looking it up in the current active recipe, or provide a default time (The previous 200 tick value)

I've also expanded the `TRTestContext` to work with non-powered machines by relaxing the type of `GenericMachineBlockEntity` to `MachineBaseBlockEntity`, and created tests for validating the expected behavior

Fixes #2850 

All GameTests pass using "runGametest" and "runGametestClient" gradle tasks

### DIFF
--- a/src/gametest/groovy/techreborn/test/TRTestContext.groovy
+++ b/src/gametest/groovy/techreborn/test/TRTestContext.groovy
@@ -28,7 +28,7 @@ import net.minecraft.item.ItemConvertible
 import net.minecraft.item.ItemStack
 import net.minecraft.test.TestContext
 import net.minecraft.util.math.BlockPos
-import techreborn.blockentity.machine.GenericMachineBlockEntity
+import reborncore.common.blockentity.MachineBaseBlockEntity
 import techreborn.init.TRContent
 
 class TRTestContext extends TestContext {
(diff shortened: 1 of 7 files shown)


Ayutac (commented):
Have checked the main files but not the test files.

Please change "cookTime" to "cookingTime" for proper English (both the method AND the local variable, or maybe even rename the local variable).

I don't like that getRecipe overwrites the last recipe, that is a side effect one would not expect from a method named like this. Maybe rename it to refreshRecipe?

Otherwise the main files are fine.

SimonFlapse:
> Have checked the main files but not the test files.
> 
> Please change "cookTime" to "cookingTime" for proper English (both the method AND the local variable, or maybe even rename the local variable).
> 
> I don't like that getRecipe overwrites the last recipe, that is a side effect one would not expect from a method named like this. Maybe rename it to refreshRecipe?
> 
> Otherwise the main files are fine.

Thanks for the quick review 😃

I named it `cookTime` because that was what it seemed to be called by the SmeltingRecipe (`getCookTime()`)
I'll refactor it.

Good point about the `getRecipe()` method. Refactoring that as well.

Ayutac:
Nice. I think it should be getCookingTime for SmeltingRecipe as well, but that is not for this PR to change. Once someone (prob modmuss) had looked over your groovy tests I think this can be merged.

PR #2868, opened 2022-03-25
Fix #2847 Allowing values larger than Integer.MAX_VALUE

### LINKED ISSUE #2847: Fusion Computer GUI is broken due to (long) to (int) cast
**Describe the bug**
It allows overflow in client GUI.

**Steps to Reproduce**
Place fusion reactor and increase its size

**Screenshots**
[If applicable, add screenshots to help explain your problem.]

(https://github.com/TechReborn/TechReborn/blob/1.18/src/main/java/techreborn/client/gui/GuiFusionReactor.java#L118)

 - Minecraft: 1.18.x
 - Mod Loader: fabric

**Logs**
Pastebin the log [e.g. `logs/latest.log`] and link it below:

**Additional context**
it should never be casted to int

SimonFlapse:
* Allowed the GUI to display `Long` numbers to fix overflow values

* Added missing Death Messages from stepping on the fusion reactor

Fixes #2847 

Old Missing translation:
image
Fixed death message:
image

Fixed Tooltip:
image

Fixes #2847

### DIFF
--- a/RebornCore/src/main/java/reborncore/client/gui/guibuilder/GuiBuilder.java
+++ b/RebornCore/src/main/java/reborncore/client/gui/guibuilder/GuiBuilder.java
@@ -564,14 +564,14 @@ public class GuiBuilder {
 	 * @param gui             {@link GuiBase} The GUI to draw on
 	 * @param x               {@code int} Top left corner where to place energy bar
 	 * @param y               {@code int} Top left corner where to place energy bar
-	 * @param energyStored    {@code int} Current amount of energy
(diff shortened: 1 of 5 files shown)


Ayutac:
Looks good. Have you checked if the method which signature you changed is used anywhere else?

SimonFlapse:
> Looks good. Have you checked if the method which signature you changed is used anywhere else?

I did a quick usage search, but it should not matter, since int -> long is an implicit cast.

SimonFlapse:
I have added a quick commit for adding the string literals to the `en_us.json` file

PR #2869, opened 2022-03-25
Added fishing junk loot

Ayutac:
Rubber and tree tap. Ah, also reordered the loot code a bit for better maintainability.

I tried to test it, but luck wasn't on my side.

After over 100 fishing attempts I gave up, this is too boring. I haven't fished leather or string either, so this doesn't tell us anything. Can someone else fish a bit? I used Lure III and constant rain.

But no compiling errors, so it probably works :)


Ayutac:
oh whoops, I thought this was merged a long time ago. Whoopsie

PR #2870, opened 2022-03-25
#2775 Adding Trinket support for BatpackItem

### LINKED ISSUE #2775: Portable energy compatibility with trinkets
Functionalities to add:
- Portable energy sources can be equipped in trinket slots
- Portable energy sources can charge items that are equipped in trinket slots

SimonFlapse:
* Added support for wearing BatpackItems in the 'back' trinket slot on the chest. This frees up an additional space for regular armor.

Closes #2775 

Also added a restriction on when the BatpackItem can output power. In 1.18.1 it can output if it is in any slot in a players inventory. I propose that it should only work when in the chest slot (or Trinkets chest/back slot)

Final notice: I am unsure about how to correctly handle the dependency for the Trinkets mod. I have to make changes to the `build.gradle` file to be able to run the mod. I fear that we now have a dependency on the Trinkets mod. Feel free to suggest alternatives or commit a fix if this is a problem.

---
### Notes
~~|The closest I have gotten to it working is inserting energy into the item, but not extracting from the portable battery pack itself.~~
~~The version I have committed, throws IllegalArgumentException, when you place a BatpackItem into the trinket slot, because It cannot find it in the players inventory.~~

~~To partially fix this replace~~
~~`EnergyStorage sourceStorage = ContainerItemContext.ofPlayerSlot(player, sourceSlot).find(EnergyStorage.ITEM);`~~
~~with~~
~~`EnergyStorage sourceStorage = ContainerItemContext.withInitial(itemStack).find(EnergyStorage.ITEM);`~~

~~However this causes the BatpackItems to never lose any energy. (I just realised, I may have been to tired. It clearly says that `withInitial` does not return the mutable stack)~~

After a good nights sleep, I managed to get a working prototype

~~I would probably do some refactoring to eliminate duplicate code blocks before marking this ready for review~~

~~I submit this PR as a draft, and welcome suggestions or further commits that will fix the last issues.~~

Feel free to leave some feedback

### DIFF
--- a/src/main/java/techreborn/items/armor/BatpackItem.java
+++ b/src/main/java/techreborn/items/armor/BatpackItem.java
@@ -90,6 +90,12 @@ public class BatpackItem extends ArmorItem implements RcEnergyItem, Trinket {
 		if (worldIn.isClient) {
 			return;
 		}
+
+		// vanilla default for chest inventory slot
+		if (itemSlot != 2) {
+			return;
+		}
+
 		if (entityIn instanceof PlayerEntity) {
 			ItemUtils.distributePowerToInventory((PlayerEntity) entityIn, stack, tier.getMaxOutput());
 		}
--- a/src/main/java/techreborn/items/armor/BatpackItem.java
+++ b/src/main/java/techreborn/items/armor/BatpackItem.java
@@ -24,22 +24,30 @@
 
 package techreborn.items.armor;
 
+import dev.emi.trinkets.api.SlotReference;
+import dev.emi.trinkets.api.Trinket;
+import dev.emi.trinkets.api.TrinketsApi;
+import net.fabricmc.fabric.api.transfer.v1.context.ContainerItemContext;
+import net.fabricmc.loader.api.FabricLoader;
 import net.minecraft.entity.Entity;
 import net.minecraft.entity.EquipmentSlot;
+import net.minecraft.entity.LivingEntity;
 import net.minecraft.entity.player.PlayerEntity;
 import net.minecraft.item.ArmorItem;
 import net.minecraft.item.ArmorMaterial;
 import net.minecraft.item.ItemGroup;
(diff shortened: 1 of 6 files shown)


src/main/resources/data/trinkets/entities/techreborn.json
```diff
@@ -0,0 +1,8 @@
+{
+	"entities": [
+		"player"
+	],
+	"slots": [
+		"chest/back"
```
Ayutac, line 6: Is that a usual trinket slot or did you make it up? Because a "/" in the name looks not so good since it is a directory separator

SimonFlapse: It is one of the default trinket slots build-in for the trinkets mod.

If only Tech Reborn and Trinkets are installed, this enables the back trinket when hovering over the chest inventory slot.

Reference:
https://github.com/emilyploszaj/trinkets/wiki/Default-Slots

src/main/java/techreborn/items/armor/BatpackItem.java
```diff
@@ -104,4 +137,15 @@ public boolean isItemBarVisible(ItemStack stack) {
 	public int getItemBarStep(ItemStack stack) {
 		return ItemUtils.getPowerForDurabilityBar(stack);
 	}
+
+	private void registerTrinket() {
+		// Ensures that trinkets are only registered if the mod is installed
+		try {
+			Class.forName( "dev.emi.trinkets.api.TrinketsApi" );
```
Ayutac: Instead use ModLoader.isLoaded(String) or the other method. Loaded is later I think and present or something sooner. Should take the sooner one

SimonFlapse: I found another check for `if the mod existed` in the code base and just used that.

PR #2874, opened 2022-03-26
Fix #2872 Converted Nickel Dust advancement to Nugget

### LINKED ISSUE #2872: Nickeldust advancement still exists
The advancement wasn't changed when we switched from dusts to nuggets. Easy to fix, but I don't have my IDE open rn.

SimonFlapse:
Fixes #2872 

Screenshot:
image


Ayutac (approved):
Excellent, thanks!

Can be merged unless there is some issue with changing advancements, which I wouldn't know about.

PR #2876, opened 2022-03-27
Added Block Breaker Machine (Adds #2871)

SimonFlapse:
# Block Breaker and Block Placer machines

This PR adds two new machine blocks in tier 0

### Block Breaker
Ability to break the block in front of it, given power and placing it inside its internal storage

### Block Placer
Ability to place a block from its internal storage, given power and a free block in front of it

### More info
Currently it takes 100 ticks to mine all block at a cost of 5 energy per tick

- [x] I would most likely be adding a Block Placer to complement this machine.
- [x] I would also like to change the time to mine blocks, based on their "hardness"
- [x] I would also like to enable redstone control for enabling/disabling the block breaking
- [x] Support upgrades (Supports most energy based upgrades. The overclock upgrade needs some more work)
- [x] Support custom configuration of values from a config file
- [x] Sound when working
- [x] Add block breaking visual progress, like when a player is mining a block
- [ ] Might need some new machine assets, I've created a quick new (temporary?) front for this machine

## Discussion:
* Should this machine effectively have silk touch? Currently it places the block it removes from the world into its inventory. However we may want it to say return a dirt block instead of a grass block when breaking a grass block.
* How much energy should it cost to break a block/per tick
* How much energy should it store (Currently set at 1000 E)
* What should the input rate for power be (Currently 32 E/t)?

## Pictures:
image

### Block Breaker:
image
image
image

### Block Placer:
image
image

~~Submitting as draft for some feedback before investing more energy on this block~~

## Known limitations (No plans to support it):
* Cannot be faced up or downwards
* Can not place fluids from buckets

## Tasks identified by reviewers

- [x] Fix configurable values set directly on the attribute
- [x] Implement dropping inventories of storage blocks (e.g. chests should not lose content when broken)
- [x] Fix exception control flow anti-pattern


SimonFlapse:
I've finished the two machines in the latest commit.

You can use the configuration files to change the amount of energy storage, transfer rate and cost to break a block.

The Block Breaker still works as if it has `silk touch`, making it return `regular stone` when breaking `regular stone` and not `cobblestone` like when a player destroys a `regular stone` without `silk touch`

SimonFlapse:
I forgot to add some recipes:
image
image

SimonFlapse:
Do we want the block breaker to drop the content of any storage block on break?

Currently if you have the block breaker, break a chest, it will delete everything contained within the chest

src/main/java/techreborn/blockentity/machine/tier0/block/BlockBreakerBlockEntity.java
```diff
@@ -0,0 +1,138 @@
...
+ * <br>
+ * May not break blocks with hardness < 0, may not break blocks that has no item.
+ * Shares most functionality with other GenericMachineBlockEntities, however does not make use of the {@code RecipeCrafter}
+ *
+ * @author SimonFlapse
+ */
+public class BlockBreakerBlockEntity extends GenericMachineBlockEntity implements BuiltScreenHandlerProvider, BlockProcessable {
+
+	public static final int BASE_BREAK_TIME = TechRebornConfig.blockBreakerBaseBreakTime;
+	public static final int COST_TO_BREAK = TechRebornConfig.blockBreakerEnergyPerTick;
```
modmuss50: Be careful doing this, as the config values may be loaded (via reflection) after this value has been set. I would just use the `TechRebornConfig.blockBreakerBaseBreakTime` directly.

I expect this may be an issue in other places, dont worry about those for now.

src/main/java/techreborn/blockentity/machine/tier0/block/BlockPlacerBlockEntity.java
```diff
@@ -0,0 +1,138 @@
...
+ * <br>
+ * May not place blocks with hardness < 0.
+ * Shares most functionality with other GenericMachineBlockEntities, however does not make use of the {@code RecipeCrafter}
+ *
+ * @author SimonFlapse
+ */
+public class BlockPlacerBlockEntity extends GenericMachineBlockEntity implements BuiltScreenHandlerProvider, BlockProcessable {
+
+	public static final int BASE_BREAK_TIME = TechRebornConfig.blockPlacerBaseBreakTime;
+	public static final int COST_TO_BREAK = TechRebornConfig.blockPlacerEnergyPerTick;
```
modmuss50: Again same nit pick.

src/main/java/techreborn/blockentity/machine/tier0/block/blockbreaker/BlockBreakerProcessor.java
```diff
@@ -0,0 +1,207 @@
...
+		} catch (MachineProcessException ex) {
+			status = ex.getStatus();
+		}
+
+		return status;
+	}
+
+	private void ensureRedstoneEnabled() {
+		if (!processable.isActive(RedstoneConfiguration.RECIPE_PROCESSING)) {
+			throw new MachineProcessException(BlockBreakerStatus.IDLE_PAUSED);
```
modmuss50: Using exceptions for control logic can be a cause for peformance issues, as when the exception is created it generates the stack trace.

See:

https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/lang/Throwable.java#L271
and
https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/lang/Throwable.java#L794


A way to solve this is to override `fillInStackTrace` in MachineProcessException

```java
@Override
public synchronized Throwable fillInStackTrace() {
    return this;
}
```

Using Exception's for control flow isnt generally a good practice in software dev, I dont think its worth changing here though, when you have the above change.

modmuss50 (commented):
Looks awesome, got a few things to point out, but nothing major 👍 

Thanks a lot.

modmuss50:
> Do we want the block breaker to drop the content of any storage block on break?
> 
> Currently if you have the block breaker, break a chest, it will delete everything contained within the chest

I believe other mods that I know of drop the content on the floor, deleting items is something id try not to do.

src/main/java/techreborn/blockentity/machine/tier0/block/blockbreaker/BlockBreakerProcessor.java
SimonFlapse: Good point, I read up on it and I agree it is an anti-pattern. (You learn something new everyday) 

I will try to change the control flow to not use exceptions, otherwise I'll make use of your suggestion.

src/main/java/techreborn/blockentity/machine/tier0/block/BlockBreakerBlockEntity.java
SimonFlapse: Thanks for pointing it out.

It was just the quickest way to support configurable values but I see the issue.

I'll do the refactoring needed to fix it.

src/main/java/techreborn/blockentity/machine/tier0/block/BlockPlacerBlockEntity.java
```diff
@@ -0,0 +1,138 @@
...
+	public void tick(World world, BlockPos pos, BlockState state, MachineBaseBlockEntity blockEntity) {
+		super.tick(world, pos, state, blockEntity);
+		if (world == null || world.isClient) {
+			return;
+		}
+
+		processor.onTick(world, getFrontBlockPosition(blockEntity.getFacing(), pos));
+	}
+
+	private BlockPos getFrontBlockPosition(Direction direction, BlockPos pos) {
```
aria1th: Just being curious but what is this for? pos.offset(blockEntity.getFacing() == null? Direction.NORTH : blockEntity.getFacing()) should do the work if its checking NPE

SimonFlapse: `Pos.offset` may very well work :) I'm still getting used to he APIs.

Thanks for suggesting

SimonFlapse:
> > Do we want the block breaker to drop the content of any storage block on break?
> > Currently if you have the block breaker, break a chest, it will delete everything contained within the chest
> 
> I believe other mods that I know of drop the content on the floor, deleting items is something id try not to do.

It was a simple change, so it should now actually break blocks instead of simply removing it from the world. This means that chests etc. drops their inventory just like when a player breaks the block

SimonFlapse:
Do we want to change the texture of the machines? I did not spent a lot of time creating them, and neither do I have the artistic skills to really do anything better.

modmuss50 (approved):
LGTM, textures also look great IMO 👍

Ayutac (changes requested):
Toasts for recipes are missing, cf. 
https://github.com/TechReborn/TechReborn/blob/1.18/src/main/resources/data/techreborn/advancements/recipes/crafting_table/machine/drain.json
Suggestion: replace iron_bars in drain recipe toast with piston and sticky piston respectively

SimonFlapse:
> Toasts for recipes are missing, cf. https://github.com/TechReborn/TechReborn/blob/1.18/src/main/resources/data/techreborn/advancements/recipes/crafting_table/machine/drain.json Suggestion: replace iron_bars in drain recipe toast with piston and sticky piston respectively

I need a checklist for everything you need to do when adding a block :P 

Thanks for catching it

SimonFlapse:
I've squashed the commits to a single nice commit. There should be references to the previous commits, if anyone wants to check each of them.

This PR should be ready for final review/merge

Ayutac (changes requested):
the refined iron plate can stay if you want it to, but please remove the electric circuit from the toasts. player gets enough toasts from electric circuit already and they are usually a requirement before obtaining refined iron plates
> Suggestion: replace iron_bars in drain recipe toast with piston and sticky piston respectively

that was meant like "and dont really change anything else" :P

Ayutac (approved):
I'm satisfied now with the toasts :)
As well as with the recipes and the balancing so far, but I should mention I did not actually test this PR in MC.

LoganDark:
> image

I know I'm a bit late for this because xanthian bullied me into muting the discussion, but the contrast here (or lack thereof) really hurts my eyes.

SimonFlapse:
> > image
> 
> I know I'm a bit late for this because xanthian bullied me into muting the discussion, but the contrast here (or lack thereof) really hurts my eyes.

It looks fine on my monitor, but I can see that it may look bad on some machines (Probably more so for budget laptops) If you think it should be changed feel free to create an issue for it. Otherwise I may change it, if i remember to do it :)

LoganDark:
> Probably more so for budget laptops

Not sure if that's intentionally insulting or not (are you calling my machine a budget laptop?), this is more of an issue with WCG displays in general (although I do admit this one is configured quite badly).

Lime green on near-white isn't a very good combo anyway, any color checker would have given that a horrible contrast score.

> If you think it should be changed feel free to create an issue for it.

I will :)

Tylertron1998:
Would appreciate a config option to disable the silk-touch nature of the block breaker. It allows players to pick up things they otherwise wouldn't be able too - which is no good.

LoganDark:
> Would appreciate a config option to disable the silk-touch nature of the block breaker. It allows players to pick up things they otherwise wouldn't be able too - which is no good.

That would probably be best served as a feature request (use Discussions, maybe). This is a long-closed PR that is no longer tracked.

PR #2836, opened 2022-02-28
Prevent IllegalStateException

aria1th:
for some reason server packet can get corrupted, but its connection should not fail.

PR #2890, opened 2022-04-01
Dispenser Scrap Box related bug fix 

aria1th:
It used Itemstack object itself instead of copy, and called decrement(1), which made original recipe's item object reference air, and sent it to client as air, and caused recipeSerde exception.


aria1th:
Closes #2836

PR #2846, opened 2022-03-04
Split client and server code

modmuss50:
Waiting on some loom fixes, and a new loader build. REI seems to have issues, I didnt look into tho.

PR #2895, opened 2022-04-05
Suggestion: Added support for Patchouli Books

SimonFlapse:
* This allows the creating of in-game books for documentations
* Provided example with Block Breaker and Block Placer
* Replaces old Tech Reborn Manual

## Notes
This PR is a concept of using the Patchouli mod for adding in-game books often used by other mods for documentation.

Let me know what you think of this idea

### Example screenshots

image
image
image
image

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -64,6 +64,14 @@ repositories {
 			includeGroup "com.github.emilyploszaj"
 		}
 	}
+    maven {
+        url = 'https://maven.blamejared.com'
+        content {
+            //Patchouli
+            includeGroup "vazkii.patchouli"
+        }
+    }
+
 }
 
 def ENV = System.getenv()
@@ -197,6 +205,8 @@ dependencies {
     optionalDependency "com.github.emilyploszaj:trinkets:${project.trinkets_version}"
     disabledOptionalDependency "net.oskarstrom:DashLoader:${project.dashloader_version}"
 
+    optionalDependency "vazkii.patchouli:Patchouli:${project.patchouli_version}"
+
     // Use groovy for datagen/gametest, if you are copying this you prob dont want it.
     gametestImplementation 'org.apache.groovy:groovy:4.0.0'
     datagenImplementation 'org.apache.groovy:groovy:4.0.0'
@@ -318,6 +328,7 @@ task renameCrowdin(type: Copy, dependsOn: ['crowdin', 'cleanCrowdin']){
 import groovy.json.JsonOutput
 import groovy.json.JsonSlurper
 import groovy.util.XmlSlurper
+import org.kohsuke.github.GHReleaseBuilder
 
 task fixTranslations(dependsOn: ['renameCrowdin']) {
     description "Remove all translations that do not have an entry, ensures that minecraft falls back to EN_US over writing out an empty string"
@@ -334,7 +345,7 @@ task fixTranslations(dependsOn: ['renameCrowdin']) {
 }
 
 
-import org.kohsuke.github.GHReleaseBuilder
+import org.kohsuke.github.GitHub
 
 curseforge {
 	if (ENV.CURSEFORGE_API_KEY) {
@@ -372,9 +383,6 @@ def getBranch() {
 
 	return "unknown"
 }
-
-
-import org.kohsuke.github.GitHub
 
 task github(dependsOn: remapJar) {
 	onlyIf {
--- a/src/main/java/techreborn/items/ManualItem.java
+++ b/src/main/java/techreborn/items/ManualItem.java
@@ -24,17 +24,19 @@
 
 package techreborn.items;
 
+import net.minecraft.client.item.TooltipContext;
 import net.minecraft.entity.player.PlayerEntity;
 import net.minecraft.item.Item;
 import net.minecraft.item.ItemStack;
 import net.minecraft.server.network.ServerPlayerEntity;
-import net.minecraft.util.ActionResult;
-import net.minecraft.util.Hand;
-import net.minecraft.util.TypedActionResult;
+import net.minecraft.text.Text;
+import net.minecraft.text.TranslatableText;
+import net.minecraft.util.*;
 import net.minecraft.world.World;
-import reborncore.common.network.NetworkManager;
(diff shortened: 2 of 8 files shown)


build.gradle
```diff
@@ -318,6 +328,7 @@ task renameCrowdin(type: Copy, dependsOn: ['crowdin', 'cleanCrowdin']){
 import groovy.json.JsonOutput
 import groovy.json.JsonSlurper
 import groovy.util.XmlSlurper
+import org.kohsuke.github.GHReleaseBuilder
```
modmuss50, line 331: Please dont make unrelated changes.

src/main/java/techreborn/items/ManualItem.java
```diff
@@ -45,9 +47,15 @@ public ManualItem() {
 	@Override
 	public TypedActionResult<ItemStack> use(final World world, final PlayerEntity player, final Hand hand) {
 		if (player instanceof ServerPlayerEntity serverPlayerEntity) {
-			NetworkManager.sendToPlayer(ClientboundPackets.createPacketOpenManual(), serverPlayerEntity);
+			PatchouliAPI.get().openBookGUI(serverPlayerEntity, new Identifier("techreborn:techreborn"));
```
modmuss50, line 50: This will crash when patchouli is not present.

This should send a packet as I expect `openBookGUI` is client only, this breaks the refactors I made to help work towards: https://github.com/TechReborn/TechReborn/pull/2846

src/main/resources/assets/techreborn/lang/en_us.json
```diff
@@ -1106,5 +1107,13 @@
...
-  "gui.techreborn.block.progress.paused": "Idle (%s)"
+  "gui.techreborn.block.progress.paused": "Idle (%s)",
+
+  "_comment27": "Patchouli Book",
+  "patchouli.techreborn.guide.name": "Tech Reborn Manual",
+  "patchouli.techreborn.landing_text": "Welcome to the world of Tech Reborn.$(br2)Check out our online $(l:https://wiki.techreborn.ovh/doku.php)Wiki$()$(br2)You can also join our $(l:https://discord.gg/teamreborn)Discord!$()$(br2)If you have feedback, suggestions, issues or want to contribute visit our $(l:https://github.com/TechReborn/TechReborn/issues)GitHub$()$(br2)We hope you enjoy playing!",
+
+  "_comment28": "Guide Book Machines",
+  "patchouli.techreborn.machine.block_breaker": "The block breaker is a simple machine. It allows you to break a block from the location in front of it, to its inventory. This action costs energy",
+  "patchouli.techreborn.machine.block_placer": "The block placer is a simple machine. It allows you to place a block from its inventory, to the location in front of it. This action costs energy"
```
modmuss50, line 1118: Im unsure if this is the best way to document TR, the wiki is the go-to place for documentation, im unsure if we want to fragment it even more?

modmuss50 (changes requested):
Some issues with this, and im unsure if this is something we even need/want as it will fragment the documentation across multiplue places.

build.gradle
SimonFlapse: Hmm weird I do not remember adding that.

src/main/java/techreborn/items/ManualItem.java
SimonFlapse: I'll look into fixing this if this concept is something we want

Ayutac (commented):
Please close this PR. Only a completed Patchouli book should be added. This is already in the works. I added you on the discord thread.

SimonFlapse:
> Some issues with this, and im unsure if this is something we even need/want as it will fragment the documentation across multiplue places.

I have experienced multiple people being very annoyed trying to craft the book, for it to just not offer any information. It feels like people are starting to get accustomed to a fully fledged reference guide in game. Furthermore this will closely link the documentation to the actual code (Wikis often become outdated, (or updated beyond the version a player is using) but an included book is expected to be up-to-date for the version it is released with.

I should clarify, I do not expect this to be merged in its current state, I'll convert it to a draft. If this is something that is interesting, this would serve as a starting point.

src/main/resources/assets/techreborn/lang/en_us.json
Ayutac: This is totally wrong. Patchouli is worked on per translation. Ie rather than adding translation keys, you add a complete book in the correct language, which is explained on the patchouli book page as well.

Regarding the wiki, it is bad insofa that it is not being kept up to date. I want to do something with datagen to automatically keep the wiki and the patchouli up to date and synchronized

Ayutac:
a clean smallscale reference implementation can be found here

https://github.com/Ayutac/time/commit/24264bd9d86e3cd2684f4a4f5146dd4f2f471cb7

src/main/resources/assets/techreborn/lang/en_us.json
SimonFlapse: Patchouli has two options for i18n

I opted for the optional Lang files support to allow the already working crowd source translation to work instead of having a new book for each language. (As suggested by the mod).

Ayutac: Okay, I admit I didn't know that.
Nevertheless, this was makes maintaining the book much harder imo

SimonFlapse: I agree, it is harder to create the book this way, but on the other hand, the alternative fragments the way translations are handled. I wanted to spark discussion.

Ayutac:
Let's do it properly. A good patchouli guide takes takes time and work and shouldn't hang on a single person's dev branch. I opened https://github.com/TechReborn/TechReborn/tree/feature/patchouli, lets work there with real PRs instead of a single draft PR onto dev

SimonFlapse:
> Let's do it properly. A good patchouli guide takes takes time and work and shouldn't hang on a single person's dev branch. I opened https://github.com/TechReborn/TechReborn/tree/feature/patchouli, lets work there with real PRs instead of a single draft PR onto dev

Sounds like a good idea :)

PR #2928, opened 2022-04-27
Added the muffler upgrade

Ayutac:
Added the muffler upgrade. The muffler upgrade makes a specific machine not play a sound. I know all sounds can be disabled via config, and if so, this upgrade becomes worthless. However, in case sounds are generally enabled because in principle you think the machine should make sounds, but maybe not directly next to your bed, the muffler upgrade is for you.

Fully tested. Review from main dev requested for how this fits the mod. The texture provided is not as good as the other upgrade textures, but may serve as a substitute until an artist has time for a high-quality one.


modmuss50 (approved):
I like it 👍

PR #2931, opened 2022-04-28
Added two TR villagers

Ayutac:
The tinkerer, that deals with alloy-related items, and the electrician, that deals with cable/battery-related items. Both sell basic and advanced circuits.

Tested, works. However, the tinkerer doesn't have a texture yet, placeholder texture is that of the armorer.

Rn these both won't spawn in villages naturally, but their job blocks are the iron alloy smelter and the (solid fuel) generator, respectively.


modmuss50 (approved):
Great start, can get a texture and possibly add world gen for these later. 👍

PR #2935, opened 2022-04-30
Fix Fluid Tanks generating static lag from trying to pull /push fluid from Items.AIR

aria1th:
Just added isEmpty() check
I expect inventory is not null, so please tell me if it should expect some weird Nullable cases?


Ayutac:
> import net.fabricmc.fabric.api.transfer.v1.item.InventoryStorage;

is used in the method, so I guess a non-null check is sensible.

PR #2938, opened 2022-05-01
Fix KIBE Entangled Bucket bug & Creative tank is creative

aria1th:
This has 3 fixes:
1. Move weird entangled buckets after transfer, Kibe's entangled bucket is moved to output slot after drain / fill if it can.
2. check isContainer condition before transfer.
3. Separate creative tank fill from item transfer, and make it actually creative

Fixes this : (low frame rate but its changing inventory rapidly)
ezgif-2-baa0f78616
'

### DIFF
--- a/src/main/java/techreborn/blockentity/storage/fluid/TankUnitBaseBlockEntity.java
+++ b/src/main/java/techreborn/blockentity/storage/fluid/TankUnitBaseBlockEntity.java
@@ -108,17 +108,27 @@ public class TankUnitBaseBlockEntity extends MachineBaseBlockEntity implements I
 			return;
 		}
 
-		if ((canDrainTransfer() && FluidUtils.drainContainers(tank, inventory, 0, 1))
-				|| FluidUtils.fillContainers(tank, inventory, 0, 1)) {
-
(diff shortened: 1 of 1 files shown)


Technici4n:
What if a very large tank is placed in the input? Ideally it should be buffered until it's empty.

Ayutac:
Can you add a concrete description of the bug? Because I can gleam a bit from the gif, but it would be better if you told us what is actually happening

aria1th:
@Technici4n It does not drain from tank item(tech reborn) but drains from MI tanks. It works correctly with MI tanks, this bug is only related to KIBE afaik.

Tech reborn tank : does not offer fluid storage, so drain / fill does not happen
MI tank (or other) : Offers fluid storage, so drain / fill happens, then it moves to output slot correctly.
Kibe bucket : it offers external fluid storage, drain / fill happens BUT it does not move to output slot somehow.

In original code, since it does not move entangled bucket to output slot, **tank repeated fill / drain again and again.** this code tries to fix it.

Technici4n:
Ah yes I see. I would check whether the tank is now full after filling / empty after emptying to decide to move. Though to be fair ender buckets are hacky, so it's also fine not to do anything about them.

PR #2943, opened 2022-05-02
Fixed the chainsaw bug

Ayutac:
For the automatic function, the chainsaw would never check the first block to be a log or leaves. Meaning you could mine e.g. a Crafting table in front of a decorative log and both the table + all the logs connected to it would be mined, if set to active. This PR fixes that.

Untested.


Ayutac:
Doesnt work because the start block in post mine is always air

PR #2955, opened 2022-05-20
Added scrap to fishing junk

Ayutac:
Changes made from phone, so untested and pls double check that it is working


drcrazy:
@Ayutac plz, update to 1.19

Ayutac:
I will tomorrow

Ayutac:
The problem is bigger than anticipted. The old loot additions were never available, since I added the code at the wrong place. And fabric doesn't seem to support loot table changes as needed for fishing. Ignore this PR for now.

PR #2960, opened 2022-05-24
fixes #2950

### LINKED ISSUE #2950: Rock cutter isn't displayed like the other handheld items
Instead it is held like a cable. Not sure why

TR 5.2.0-beta.3, MC 1.18.2

Ayutac:
Needs to be tested if it is now displayed correctly. 1.19 version of #2956


Ayutac:
Before the fix:


After the fix:

drcrazy:
It looks like big cylinder is a storage for sap while small handle below it  - is a handle )

PR #2962, opened 2022-05-24
Recipes for 1.19

Ayutac:
I'm a bit slow on this, sorry


Ayutac:
It's not complete imo, but it contains the important stuff. Maybe I will add more later, but this can be merged if 1.19 comes out, after you test it of course.

Ayutac:
tested, works, and I am satisfied for now :)

PR #2964, opened 2022-05-25
Added ind. loot to ancient city

Ayutac:
I left the ice box out for now, maybe some other time.

Done via web editor, needs testing!


Ayutac:
tested, works, and actually makes the ancient city feel worth the danger :D

PR #2968, opened 2022-05-30
Add silicon to the fluid replicator's recipes 

WaffleDevsAlt:
I think this is all thats needed, never made a mod before so i probably forgot something.


WaffleDevsAlt:
Tested and works
image

PR #2972, opened 2022-06-02
Added rutile ore

Ayutac:
Tested, works. Requires datagen rerun. Second pair of eyes to make sure I didn't forget anything.

Currently the three ores can only be grinded down to dust, but not further changed to titanium dust. This is intentional, as this is a step-by-step work. 

I would recommend merging this PR anyway for the world generation aspect.


Ayutac:
@drcrazy said we have enough ores as it is, but what if we defaulted generating this ore to false?

drcrazy:
There is a point to not have nickel and titan. They do require some study and automation. Otherwise it will be more like a regular metals

Ayutac:
rutile would require quite some automaization to get titanium out of it. I will be back when I can illustrate the whole process

BernardoGrigiastro:
Not intended to sound rude at all but the point to not have nickel is to punish players? Imho its a major pita, while testing 1.19.3 TR the last days I keep cursing it and thinking on making an ultra simple mod just to add nickel and zinc ores (launching intellij as we speak..). I kinda agree on titan but the other two are basic materials that are only "hard" when TR is the only mod in and  kinda goes against the villagers thing, they are making TR finally easier, thank god (and/or ayutac :D )

PR #2987, opened 2022-06-19
feat(FusionReactor): add check for redstone signal for crafting

### LINKED ISSUE #2986: Fusion Control Computer ignores the setting to disable crafting when a redstone signal is present
**Describe the bug**
The fusion control computer ignores the setting to disable crafting when a redstone signal is present.

**Steps to Reproduce**
1. Build the simplest fusion reactor.
2. Charge it, let it run.
3. Set `Crafting` to any value other than `Ignored`
4. Try to turn it off (by not providing a redstone signal when the GUI is set to `Redstone ON` or vice versa).
5. The fusion control computer ignores the setting and continues crafting.

**Screenshots**

 - Minecraft: 1.18.2
 - Mod Loader: Fabric 0.14.8

M0dEx:
This PR adds logic for pausing the fusion reactor depending on the redstone `Crafting` setting.

This logic is missing for multiple machines that should have it - industrial electrolyzer, for example.

I also added a translatable string, so if this gets merged, it would be nice to update Crowdin.

Fixes #2986.

### DIFF
--- a/src/main/java/techreborn/blockentity/machine/multiblock/FusionControlComputer/FusionControlComputerBlockEntity.java
+++ b/src/main/java/techreborn/blockentity/machine/multiblock/FusionControlComputer/FusionControlComputerBlockEntity.java
@@ -107,6 +107,32 @@ public class FusionControlComputerBlockEntity extends GenericMachineBlockEntity 
 	}
 
 	/**
+	 * Updates this.state according to the overall state of the reactor
+	 *
+	 * @return True if the machine is doing something, False if not.
+	 */
+	private boolean updateState() {
+		this.state = FusionControlComputerState.INACTIVE;
+		
+		if (!isMultiblockValid())
+			return false;
+		
+		if (this.state != FusionControlComputerState.CHARGING && !isActive(RedstoneConfiguration.RECIPE_PROCESSING)) {
+			this.state = FusionControlComputerState.PAUSED;
+			return false;
+		}
+		
+		if (currentRecipe == null)
+			this.state = FusionControlComputerState.NO_RECIPE;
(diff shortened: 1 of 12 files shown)


src/main/java/techreborn/blockentity/machine/multiblock/FusionControlComputer/FusionControlComputerBlockEntity.java
```diff
@@ -22,7 +22,7 @@
  * SOFTWARE.
  */
 
-package techreborn.blockentity.machine.multiblock;
+package techreborn.blockentity.machine.multiblock.FusionControlComputer;
```
Technici4n: Why?

M0dEx: So it's in a different package together with the State Enum.

The alternative would be to have the enum in the class file (`FusionControlComputerBlockEntity`). 
I don't like it, but if you would prefer it, I can move it.

Technici4n: You can keep it in another package, but please make it follow java conventions: `fusion` will do as the package name.

Technici4n: Actually, it would probably be better to make it a nested enum called `State` (in the same file), given that it's not used in many other places.

M0dEx: Fixed.

PR #2990, opened 2022-06-20
Fix a few alloy recipes

ghost:
- Made brass ratio consistent with Create
- Centrifuge recipes for alloy dusts now produce correct ratios

### DIFF
--- a/src/main/resources/data/techreborn/recipes/centrifuge/brass_dust.json
+++ b/src/main/resources/data/techreborn/recipes/centrifuge/brass_dust.json
@@ -1,20 +1,21 @@
 {
   "type": "techreborn:centrifuge",
   "power": 10,
-  "time": 1500,
+  "time": 6000,
   "ingredients": [
     {
-      "tag": "c:brass_dusts"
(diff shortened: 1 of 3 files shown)


Ayutac:
1. Why do we need to have same ratio as Create?
2. "Correct ratios" source?

drcrazy:
3 Copper + 1 Zinc -> 4 brass  is what we used to since GT4 https://ftbwiki.org/Brass_Ingot_(GregTech_4)
Centrifuge was never meant as uncrafting method with 100% yield.

goldenapple3:
1. Fair enough, I can just change the few recipes for Brass in my own modpack
2. @drcrazy the current centrifuge recipes already give 100% material yield but it's the wrong materials. 1/2 silver + 1/2 gold turns into 5/9 gold and 4/9 silver. This allows "transmuting" silver into gold (albeit very slowly), this is an exploit which does not require any other mods installed.

drcrazy:
Sounds reasonable. Can you change PR to include only #2, please?

goldenapple3:
Fixed

PR #3015, opened 2022-08-01
added techreborn:ores to c:ores

Ayutac:


drcrazy:
Should it be blocks rather then items? I suspect Diggus Maximus check block tags to do a vein mining.

Ayutac:
that would require a bit more work. I can look into it after my TR compiles again

drcrazy:
@Ayutac Wanna complete this?

src/datagen/groovy/techreborn/datagen/tags/TRBlockTagProvider.groovy
```diff
@@ -48,5 +52,9 @@ class TRBlockTagProvider extends FabricTagProvider.BlockTagProvider {
 			.addOptionalTag(BlockTags.AXE_MINEABLE.id())
 			.addOptionalTag(FabricMineableTags.SHEARS_MINEABLE.id())
 			.addOptionalTag(FabricMineableTags.SWORD_MINEABLE.id())
+
+		TRContent.Ores.values().each { ore ->
+			getOrCreateTagBuilder(TagKey.of(Registry.BLOCK_KEY, new Identifier("c", "ores"))).add(ore.asBlock())
+		}
```
velna, line 58:
```	
	TRContent.Ores.values().each {
			getOrCreateTagBuilder(ConventionalBlockTags.ORES)
				.add(it.block)
	}
```
Will do the trick.

velna:
> Should it be blocks rather then items? I suspect Diggus Maximus check block tags to do a vein mining.

This patch will also fix the problem that the Jackhammer breaks TR ore blocks.

drcrazy:
PR Obsolete. BlockTags are Implemented in TR datagen.

PR #3020, opened 2022-08-12
add missing zombie villager textures

Tomatobird8:
no more missing texture villager zombies🤠


Ayutac:
The whole time I was wondering if we can simply copy the assets or if zombie textures are different?

PR #3054, opened 2022-11-07
added GPS by Ayutac

Ayutac:
crafted with basic display over a compass. use to show coordinates.

Tested, works.


src/main/java/techreborn/items/GpsItem.java
```diff
@@ -0,0 +1,86 @@
...
+				.append(Text.literal(" Z:").formatted(Formatting.GRAY))
+				.append(Text.literal(String.valueOf(pos.getZ())).formatted(Formatting.GOLD)));
+			return new TypedActionResult<>(ActionResult.SUCCESS, stack);
+		}
+		return new TypedActionResult<>(ActionResult.PASS, stack);
+	}
+
+	@Override
+	public void appendTooltip(ItemStack stack, @Nullable World worldIn, List<Text> tooltip, TooltipContext flagIn) {
+		getPos(stack)
```
modmuss50: Question: Where is this position wirtten to the ItemStack?

Ayutac: oops! I couldn't figure it out and intended to delete that. This has been copied from the transmitter, so it is wrong there as well I suppose. On it!

modmuss50 (approved):
Nothing wrong with this, just not too sure on how useful it may be. Quite happy to merge this but I wonder if it could be a little more intresting 🤔

Ayutac:
you are free to extend on it ;)

It is semi-useful as of now because of F3, but F3 is a debug tool and coordinates are important enough that they should be available ingame without F3 imo. And since GPS is tech and this is TR, I thought it would be a neat little addition :)

PR #3060, opened 2022-11-10
1.19 launchpad and elevator, textures by Spearkiller

Ayutac:
Launchpad and Elevator for TR. Whats missing is the mixin for the player jumping/sneaking to activate it, that should call teleportUp/Down on the ElevatorBlockEntity.

(Note that the Launchpad is also contained in this PR rn)


src/main/java/techreborn/blockentity/machine/tier1/ElevatorBlockEntity.java
```diff
@@ -0,0 +1,201 @@
...
+import techreborn.init.TRBlockEntities;
+import techreborn.init.TRContent;
+
+import java.util.List;
+import java.util.Optional;
+
+public class ElevatorBlockEntity extends PowerAcceptorBlockEntity implements IToolDrop, BuiltScreenHandlerProvider {
+
+	public static final int MIN_HEIGHT = -64;
+	public static final int MAX_HEIGHT = 319;
```
modmuss50: You should use `world.isOutOfHeightLimit()` or `getBottomY()` / `getTopY()` as diffrent worlds can have different heights

src/main/java/techreborn/blockentity/machine/tier1/ElevatorBlockEntity.java
```diff
@@ -0,0 +1,201 @@
...
+
+	public ElevatorBlockEntity(BlockPos pos, BlockState state) {
+		super(TRBlockEntities.ELEVATOR, pos, state);
+	}
+
+	public static boolean isRunning(final World world, final BlockPos pos) {
+		final BlockEntity entity = world.getBlockEntity(pos);
+		if (!(entity instanceof ElevatorBlockEntity)) {
+			return false;
+		}
```
modmuss50: Dont make this a static method, removes the need for a 2nd getBlockEntity() call.

src/main/java/techreborn/blockentity/machine/tier1/ElevatorBlockEntity.java
```diff
@@ -0,0 +1,201 @@
...
+		return ((ElevatorBlockEntity)entity).getStored() > 0;
+	}
+
+	public static boolean isFree(final World world, final BlockPos pos) {
```
modmuss50: No need for this to be static.

src/main/java/techreborn/blockentity/machine/tier1/ElevatorBlockEntity.java
```diff
@@ -0,0 +1,201 @@
...
+		return world.getBlockState(pos.up()).isAir() && world.getBlockState(pos.up().up()).isAir();
+	}
+
+	public static boolean isValidTarget(final World world, final BlockPos pos) {
```
modmuss50: No need for this to be static. Et al.

src/main/java/techreborn/blockentity/machine/tier1/ElevatorBlockEntity.java
```diff
@@ -0,0 +1,201 @@
...
+		return Math.min(Math.abs(endPos.getY()-startPos.getY())*TechRebornConfig.elevatorEnergyPerBlock,0);
+	}
+
+	protected boolean teleport(World world, BlockPos pos, PlayerEntity player, BlockPos targetPos) {
+		final int energy = energyCost(pos, targetPos);
+		if (getStored() < energy) {
+			return false;
+		}
+		world.playSound(null, pos, SoundEvents.ENTITY_ENDERMAN_TELEPORT, SoundCategory.BLOCKS, 1f, 1f);
+		player.teleport(targetPos.getX(), targetPos.getY(), targetPos.getZ());
```
modmuss50: Use FabricDimensions.teleport, its much better.

src/main/java/techreborn/blockentity/machine/tier1/ElevatorBlockEntity.java
```diff
@@ -0,0 +1,201 @@
...
+			return Optional.of(downPos);
+		}
+		return Optional.empty();
+	}
+
+	public static int energyCost(BlockPos startPos, BlockPos endPos) {
+		return Math.min(Math.abs(endPos.getY()-startPos.getY())*TechRebornConfig.elevatorEnergyPerBlock,0);
+	}
+
+	protected boolean teleport(World world, BlockPos pos, PlayerEntity player, BlockPos targetPos) {
```
modmuss50: No need for the World world param, you can get it from the BE.

src/main/java/techreborn/blockentity/machine/tier1/ElevatorBlockEntity.java
```diff
@@ -0,0 +1,201 @@
...
+
+	// PowerAcceptorBlockEntity
+	@Override
+	public void tick(World world, BlockPos pos, BlockState state, MachineBaseBlockEntity blockEntity) {
+		super.tick(world, pos, state, blockEntity);
+		if (world == null || getStored() <= 0 || !isActive(RedstoneConfiguration.POWER_IO)) {
+			return;
+		}
+
+		Optional<BlockPos> upTarget = null;
```
modmuss50: I dont think you need both of these, you can just have target. Will let you simplify the code a bit.

modmuss50 (changes requested):

src/main/java/techreborn/blockentity/machine/tier1/ElevatorBlockEntity.java
Ayutac: nope, this is to check if another elevator is running when calculating the target from the current elevator

src/main/java/techreborn/blockentity/machine/tier1/ElevatorBlockEntity.java
Ayutac: the idea was when mutliple players teleport at the same time, to only calculate their target once. still holds for sneaking, has been made redundant for jumping

src/main/java/techreborn/blockentity/machine/tier1/ElevatorBlockEntity.java
Ayutac: I was looking for those, thanks!

src/main/java/techreborn/packets/ServerboundPackets.java
```diff
@@ -171,6 +174,29 @@ public static void init() {
...
+
+			server.execute(() -> {
+				BlockEntity blockEntity = player.world.getBlockEntity(pos);
+				if (blockEntity instanceof LaunchpadBlockEntity) {
+					((LaunchpadBlockEntity) blockEntity).handleGuiInputFromClient(buttonAmount);
+				}
+			});
+		}));
+
+		NetworkManager.registerServerBoundHandler(new Identifier("reborncore", "jump"), (server, player, handler, buf, responseSender) -> {
```
modmuss50: You shouldnt listen to other mods packets like this. The packet should be sent from TR and recieved by TR

src/main/java/techreborn/packets/ServerboundPackets.java
```diff
@@ -171,6 +174,29 @@ public static void init() {
...
+			BlockPos pos = buf.readBlockPos();
+
+			server.execute(() -> {
+				BlockEntity blockEntity = player.getWorld().getBlockEntity(pos.down());
+				if (blockEntity instanceof ElevatorBlockEntity elevator) {
+					elevator.teleportUp(player.getWorld(), pos.down(), player);
```
modmuss50: You should check that the player is within a few blocks of the evevator, as right now someone could easily send this packet from anywhere in the world and get teleported.

modmuss50 (changes requested):

src/main/java/techreborn/packets/ServerboundPackets.java
Ayutac: But the sending happens in the mixin? How am i supposed to have the mixin in core and the sending in TR?

src/main/java/techreborn/packets/ServerboundPackets.java
Ayutac: doesnt the code require that the block below the player is an elevator?

src/main/java/techreborn/packets/ServerboundPackets.java
modmuss50: You should create an event in RC, RC should not be sending jump packets.

src/main/java/techreborn/packets/ServerboundPackets.java
modmuss50: Doesnt look like the code does. You cannot trust any packet data coming from the client.

src/main/java/techreborn/packets/ServerboundPackets.java
Ayutac: I'm out of my depth at this point then.

src/main/java/techreborn/packets/ServerboundPackets.java
Ayutac: Ah, I see my mistake now. But I would wait with fixing until the "event" is properly implemented

modmuss50 (approved):
Awesome 👍

PR #3157, opened 2023-07-06
Fluid buckets place fluid from dispenser

WaffleDevs:
Dispensing a bucket now places and collects TR added fluids.
https://github.com/TechReborn/TechReborn/discussions/3156


WaffleDevs:
Whoops. Reworking my fork structure. PR will be back soon.

Issue #3039, opened 2022-09-13
1.18.2 "Dupe" Glitch Crude Storage Unit


DarkoroDragon:
**Describe the bug**
A locked crude storage unit set to auto input from an adjacent container can pull in items that are not locked to the crude storage unit.  Manually inserting a locked item into that crude storage unit turns all of the pulled items into the locked item.

**Steps to Reproduce**

1. Place a crude storage unit next to another container (confirmed with barrels and hoppers that dont point directly into the storage unit).
2. Insert an item into the crude storage unit and lock it.  (confirmed with feathers).
3. Remove the item from the storage unit so that it is empty.  Confirm that the item is still locked onto the storage unit.
4. Turn on Auto-Input for the storage unit, facing the adjacent container.
5. Input a non-locked item into the adjacent container (confirmed with eggs).
6. The non-locked item should be pulled into the storage container.
7. Place a locked item (in the confirmed example, a feather) manually into the input slot of the storage unit.
8. The non-locked item (the egg) should be replaced by 2 of the locked item (2 feathers).

**Screenshots**
![image](https://user-images.githubusercontent.com/113436945/189890664-f22fb504-d4a1-4cca-92cc-0fd9c29dab6a.png)
![image](https://user-images.githubusercontent.com/113436945/189890708-e2703af3-f6b0-4fff-ba5e-28491094ba7d.png)
![image](https://user-images.githubusercontent.com/113436945/189890746-4d43a8b3-90a8-46cb-ab48-54e8b7414b67.png)

Note the 16 eggs, and the 43 feathers, becomes 59 feathers.

**Environment (please complete the following information with the version):**
 - Minecraft: 1.18.2
 - Mod Loader: Fabric
 - Modpack: Valhelsia Enhanced Vanilla 1.4.2
 - Tech Reborn version: 5.2.0

**Logs**
Pastebin the log [e.g. `logs/latest.log`] and link it below:
https://pastebin.com/X90SYc88

**Additional context**

DarkoroDragon:
Update: This happens if you set the auto-input of items onto the output slot with the wrench icon, but not if you only set the auto-input using the input slot on the UI.

Holy-Fak:
That is critical, someone just used this bug copy items.

drcrazy:
Sorry, no efforts from TR dev team on 1.18. Feel free to submit a PR.

PR #3171, opened 2023-07-31
 fix overwriting output slot #3039 

TGWaffles:
Fix issue #3039 where storage units could have their output slot filled with a different item, which counted towards the stored item & caused a duplication glitch.

Prevents items in the output slot from being overwritten by stored items, and also prevents them from counting when calculating quantity.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -41,7 +41,7 @@ repositories {
         }
 	}
     maven {
-        url = 'https://ladysnake.jfrog.io/artifactory/mods'
+        url = 'https://maven.ladysnake.org/releases'
         content {
             // Trinkets dependency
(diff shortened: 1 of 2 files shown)


modmuss50:
We no longer update 1.18, does this change also apply to 1.20? Thanks.

TGWaffles:
> We no longer update 1.18, does this change also apply to 1.20? Thanks.

Nope - [drcrazy said to submit a PR for the issue to be solved](https://github.com/TechReborn/TechReborn/issues/3039#issuecomment-1546752819). This leads to a duplication glitch on modpacks still on 1.18 (ie. most modpacks as many other mods are still not ported to 1.20), so quite a severe issue that I'm sure they'll appreciate being fixed.

PR #3197, opened 2023-11-05
Make Overclocker tooltips reflect reality

### LINKED ISSUE #3196: Tech Reborn Overclocker Upgrade tooltips don't reflect reality
(This is Tech Reborn 5.8.7 on Minecraft 1.20.1)

## Current state
oc
16 upgrades with Shift:
oc_shift

The issues with this:
- You can't install more than 4 overclockers in a machine at once, so stats for more than 4 overclockers are completely out-of-place,
- The overclock speed is capped at 97.5% speed increase, regardless of configs,
- The power increase is **multiplicative**, not **additive**.

## The real stats after testing machines with different amounts of overclockers:
obraz

**I made a suggestion in Discussions related to this issue - more on that [here](https://github.com/TechReborn/TechReborn/discussions/3195).**.

KonSola5:
Merging this PR will fix #3196.

This PR changes the overclocker tooltips in these ways:
- The stack info is now capped at 4 upgrades.
- Increase/Decrease color is now applied at the stat rather than the explanation, what is this stat.
- Added new calculate methods to satisfy the overclocker stat special needs.
- The overclocker "Time Decrease" is now capped at 97.5% (experimentally determined by processing a recipe taking 200 seconds with 4 overclockers)
- Stats now support doubles (instead of only ints) - rounded using `NumberFormat` to ensure that the decimal point is only shown when needed.

Also added a `SPEED_CAP` constant to the `MachineBaseBlockEntity.java` - pretty weird that the cap was supposed to be 99%, not 97.5% like it is in reality.

overclock_noshift
overclock_shift
obraz

### DIFF
--- a/RebornCore/src/main/java/reborncore/common/blockentity/MachineBaseBlockEntity.java
+++ b/RebornCore/src/main/java/reborncore/common/blockentity/MachineBaseBlockEntity.java
@@ -77,6 +77,8 @@ public class MachineBaseBlockEntity extends BlockEntity implements BlockEntityTi
 	private boolean markSync = false;
 	private int tickTime = 0;
 
+	public static double SPEED_CAP = 0.99;
+
 	/**
 	 * <p>
 	 *  This is used to change the speed of the crafting operation.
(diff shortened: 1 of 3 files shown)


KonSola5:
Of course, this works for any config:
overclocker_noshift
overclocker_shift
(here I changed the power "multiplier" to 1 and speed "multiplier" to 0.24)

(patiently waiting for 1.21 or 1.22)

PR #3212, opened 2024-01-19
Add a keybind for toggling the nano suit night vision effect

Thepigcat76:
Added:
- A keybind for toggling the nano suit night vision effect
- Handle nanosuit activity through isActive nbt tag

Changed:
- Renamed keybind translation keys to use singular instead of plurarl


src/client/java/techreborn/TechRebornClient.java
```diff
@@ -24,8 +24,10 @@
 
 package techreborn;
 
+import dev.architectury.event.events.client.ClientTickEvent;
```
drcrazy: Will it work with TR only, e.g. without REI ?

Thepigcat76: Oops, forgot to clean up unused imports

PR #3237, opened 2024-03-09
Toggle quantum legs sprint boost

justross:
Similar to toggling night vision on the nanosuit, I want the option to toggle the movement speed boost on the quantum legs as well.

I created a new keybinding with the description "Quantum suit - toggle sprint speed". It is by default bound to b, which is not a great key, but it is unconflicted with vanilla MC's keybindings.

Everything else is basically behavior I took from the nanosuit's night vision code and ported to the quantum suit.

### DIFF
--- a/src/client/java/techreborn/TechRebornClient.java
+++ b/src/client/java/techreborn/TechRebornClient.java
@@ -138,6 +138,12 @@ public class TechRebornClient implements ClientModInitializer {
 			}
 		});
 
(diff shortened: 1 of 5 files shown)


drcrazy:
@justross  Would you mind rebasing this on top of latest version?

justross:
@drcrazy Done! Thanks for reviewing!

Issue #3186, opened 2023-09-28
Quantum armor flight bug on servers only seems like


PurelyRain:
Playing tech reborn for 1.20.1 on a fabric server, I have tested quantum armor flight in both LAN and single player worlds but no matter if its a custom server hosted on my pc with a server file or a paid for one (im playing on a paid server with my friends) the flight takes a solid minute to activate, it does one of these:
lets you hover slightly by i assume turning your flight on and off because the screen becomes buggy as if the FOV keeps being changed
lets you hover a bit "better" but the FOV changing effect becomes "faster"
actually lets you fly 
now, it is random on how many times i have to press the space bar to actually fly, sometimes its first try and sometimes i have to struggle for 2 minutes to have flight properly activate, i have looked through configs and such to see if anything could be causing this but i have not gotten anywhere.
thanks for any help in advance!

drcrazy:
Do you have other mods installed besides TR?

PurelyRain:
> Do you have other mods installed besides TR?

Yes I do, I believe there were nearly 100 mods? But the thing is that the mods don't change, there's no mods that are only server side or only client side, so I don't see why they would matter when the bug persists only on servers, I even tried with just TR and it still kept happening.

drcrazy:
So it's a LAN with just a TR, right?

PurelyRain:
> So it's a LAN with just a TR, right?

Nah it's a paid server where the bug happens, LAN and single player is fine tho

drcrazy:
Then in could be something on that server which mess with either this:
https://github.com/TechReborn/TechReborn/blob/3ef8b87412086d009982e5715e0e3f2392771e89/src/main/java/techreborn/items/armor/QuantumSuitItem.java#L94

or that: 
https://github.com/TechReborn/TechReborn/blob/3ef8b87412086d009982e5715e0e3f2392771e89/src/main/java/techreborn/items/armor/QuantumSuitItem.java#L99

drcrazy:
And both of these are vanilla methods :(

Trent2087:
This is an issue with playerAbillityLib. In 1.19 and previously, PAL would just spam the console and not break flight. Now it 1.20.1 it breaks TR Quantum Armor Flight, while also spamming the console infinitely.

burdoto:
[...]

PR #3241, opened 2024-03-31
Use PlayerAbilityLib for quantum chestplate flight

top-sekret:
This is a very simple patch that adds PlayerAbilityLib support to quantum chestplate flight. I wrote it after experiencing interference with other forms of creative flights on large modpacks and have since been porting it to newer versions without any issues. Now, I have come up with the great idea to submit a PR.

This patch is pretty straightforward. Possibly the only nontrivial part of logic is that this only applies the energy cost of flight if the chestplate is **the** source of flight, due to `isActivelyGranting`. For example, in creative mode or using a sustainable flight provider (think some angel ring), this should not discharge while flying.

This probably addresses #3186.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -32,6 +32,16 @@ allprojects {
     apply plugin: "fabric-loom"
     apply plugin: "maven-publish"
 
+    repositories {
+        maven {
+            name "Ladysnake Mods"
+            url "https://maven.ladysnake.org/releases"
+            content {
+                includeGroup "io.github.ladysnake"
+            }
+        }
+    }
+
     loom {
         runtimeOnlyLog4j = true
         splitEnvironmentSourceSets()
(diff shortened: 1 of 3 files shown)


modmuss50:
Hi, thanks for the PR however I dont want to depend on 3rd party libraries, it means we have to wait for them to update amougst other issues. I will look into finding a better solution for this, as it shouldnt need a library.

averyterrel:
> Hi, thanks for the PR however I dont want to depend on 3rd party libraries, it means we have to wait for them to update amougst other issues. I will look into finding a better solution for this, as it shouldnt need a library.

Could that better solution be conditionally using PAL, when it is available? It would only be a build-time dependency then.

lexi-the-cute:
Can we please look into some way for this to be compatible with PAL, so that way flight doesn't break when PAL is installed?

PR #3247, opened 2024-05-05
Fix elevator jump 

TechPizzaDev:
Some mods (commonly Amecs) inject into KeyBinding which break ClientJumpEvent.
Amecs seems to cancel the keypress so the event never triggers.


NickAcPT (approved):
LGTM

modmuss50 (approved):

PR #3273, opened 2024-07-13
Fix #3272

### LINKED ISSUE #3272: The nano suit does not absorb any damage at all, and the nano sword does not increase damage even when activated
The nano suit does not absorb any damage at all, and the nano sword does not increase damage even when activated.

- just try nano suit and nano sword. perhaps the quantum suit is the same..

 - Minecraft: 1.21
 - Mod Loader: Fabric 0.15.11

monochromem1nd:
Work around #3272


modmuss50:
I think this is actually the fix and not a workaround, thanks for the help 👍

PR #3274, opened 2024-07-14
EMI integration

monochromem1nd:
This PR aims to add integration with another popular recipe viewer - EMI

### DIFF
--- a/RebornCore/src/main/java/reborncore/common/blockentity/MachineBaseBlockEntity.java
+++ b/RebornCore/src/main/java/reborncore/common/blockentity/MachineBaseBlockEntity.java
@@ -301,7 +301,7 @@ public class MachineBaseBlockEntity extends BlockEntity implements BlockEntityTi
 	public void writeNbt(NbtCompound tagCompound, RegistryWrapper.WrapperLookup registryLookup) {
 		super.writeNbt(tagCompound, registryLookup);
 		if (getOptionalInventory().isPresent()) {
-			getOptionalInventory().get().write(tagCompound);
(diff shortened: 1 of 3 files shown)


modmuss50:
Hi, this is a good start and is something I have had in the back of my mind. I was unsure if JEI or EMI was a better alternative. Out of intrest why did you choose EMI over JEI?

It would be nice if we could share some code between each recipe viewer mod to try and reduce the effort required to maintain them all, I am not familar at all with EMI so its not clear to me how feasible this would be. This might be a more long term goal, as having something working is likely better than being stuck in refactoring hell.

Another thing I would like to look at doing is setting up Gradle in a way to make it easy to run the game with REI or EMI, likely by having 2 run configs. Again this doesnt need to be done as part of this PR.

Do let me know if you have any questions or are stuck with anything, Id be happy to help.

monochromem1nd:
Hi. Actually creating this PR I also thought between JEI and EMI (ideally of course to have all three:)), but made an emphasis in favor of EMI due to the fact that recently there is a rapid increase in popularity of this recipe viewer and support runtime JEI him. 

Regarding code sharing - I have no idea how much will work out to maximize common ground between the Jei/Rei API and EMI.

modmuss50:
Hi, @Ftrless I have spoken to Mezz (the JEI author) and found out that EMI supports JEI plugins, so the better solution here might be to support JEI and then we get EMI support for free. Saving the need to do this all again.

monochromem1nd:
Hi @modmuss50, yeah thats probably what I was talking about just a little bit in opposite direction, so I think ure right, could reorient this pr to support for jei which would be more useful

PR #3295, opened 2024-09-14
Nickel ore

d4rkEx:
From my experience playing with this mod and watching videos, I got the impression of an imbalance due to the lack of nickel ore in the regular world.


Thepigcat76:
Its not an imbalance, its a design choice as nickel does not have many uses and therefore isn't supposed to be its own ore (tech reborn kinda has an ore bloat problem). Instead the player needs to setup infrastructure to centrifuge existing ores.

drcrazy:
Nickel produced from other ores. Moreover, there is an advancement aimed on helping players to achieve stable nickel production.

PR #3299, opened 2024-10-02
Use vanilla trade factories

DrexHD:
This reduces the amount of code required and allows my mod villagerconfig to automatically datagen trade files for these trades.

This should have no affect on the actual trades and has been validated by checking the first level of the Metallurgist


modmuss50 (approved):
Great, many thanks.

PR #3309, opened 2024-10-25
Automatic Resource Mining

LauchInterceptor:
# Goals
## Automatic Miner
A placeable tier 2 miner machine that can mine for resources automatically when supplied with power among other things.
It's supposed to be a good mid-game alternative to searching for ores manually.

### How it works
The miner needs a mining tool to function. Any of the drills or the jackhammers should work with taking thier respective block mining speeds into consideration. I've also considered including the rockcutter so the miner can silk touch but this might be too overpowered. What do you think?
The miner works much the same as the old ic2 miner does, it extends a pipe level by level and searches for ores in its influence which it then mines and outputs to an inventory or an internal buffer. This means the miner also needs mining pipe.

## Ore Scanner
A multi tiered handheld device much like the OD and OV Scanner in IC2. Scans for the presence of ores in the current location. Can be used to make the miner scan for ores instead of digging straight down.

## Balancing limitations
- The miners range is limited
- It's not very fast, when compared to something like an rf tools builder or mekanism automatic miner for instance
- It consumes a lot of power, more than manually mining the ores would.
### Optional
- Mining pipes have a chance to break on each operation

# Summary
Added Blocks
- Miner
- Mining Pipe

Added Items
- Ore Scanner (Tier 1, 2 ,3)

Please leave some feedback on what you think of the scope and goals of this project.


Thepigcat76:
ooooh very cool! Auto mining is something TR has been missing for a long time.
Only criticism is that adding 3 different tiers of ore scanners is maybe not the best idea, since typically special electrical devices only have one variant. (Only exception are drills and chainsaws cuz they are so commonly used)

LauchInterceptor:
Thank you for your feedback. I agree it would feel out of place. 
Im trying to think of a way to add multiple prospecting range upgrades to the miner though. Any ideas are very welcome :)

Thepigcat76:
Hmm the pump just has a button to increase its range iirc, but you could also look into adding a range upgrade (like the overclocker, but for prospecting instead of speed). Maybe we could also have the pump use that

LauchInterceptor:
Thats a great idea actually. I would still keep one ore scanner for adding the horizontal prospecting functionality but add the range upgrades aswell and see how it play's out. Im going to hold off on changing the range mechanic for the pump for the time being though and make it up to @maxvar to decide if he wants it/wants to implement it.

src/main/java/techreborn/blockentity/machine/tier2/MinerBlockEntity.java
```diff
@@ -0,0 +1,227 @@
...
+	static final float PROSPECTING_SPEED = 1f;
+
+	static final int MINING_TOOL_SLOT = 0;
+	static final int PROSPECTING_TOOL_SLOT = 1;
+
+	static final int MINING_PIPE_SLOT = 2;
+	static final int OUTPUT_SLOT = 3;
+	static final int ENERGY_SLOT = 4;
+
+	static final Set<Item> ALLOWED_MINING_TOOLS = Set.of(TRContent.BASIC_DRILL, TRContent.ADVANCED_DRILL, TRContent.INDUSTRIAL_DRILL);
```
drcrazy: Maybe it is better to use tags here

LauchInterceptor: Should I just define it in resources/data or add to datagen?

LauchInterceptor:
# Progress Update
The main functionality of the miner is implemented but there are a few things that still need to be done.

- [ ]  Syncing the miner status

- [ ] Slot configuration for the output inventory
- [ ] Energy consumption
- [ ] **Balancing** (with your feedback)
	- [ ] Crafting recipe
	- [ ] Overall speed and speed increase with the tools used
- [ ] **GUI** : Prettying it up (feedback is also appreciated)
	- [ ] Adding sillouettes to the tool, scanner and mining pipe slot
	- [ ] Status display
- [ ] **Art** : Finalizing the existing assets
	- [ ] mining pipes are still fences with iron/netherite block texture
	- [ ] maybe someone has a cool idea what the final miner could look like, im not an artsy person
- [ ] **Scanner**
	Im not firm on what the scanner should be but im thinking of a directional metal detector for ores kinda vibe

Thepigcat76:
Good job with this pr so far! For the art, maybe you could ask spear killer on the discord. They do all the tech reborn assets nowadays

LauchInterceptor:
Thank you for the kind words and the suggestion. I contacted him about the assets

LauchInterceptor:
This is not abandoned I just have not been able to work on it in the past couple of weeks

PR #3328, opened 2024-11-19
Fix LESU Network. Closes #3325

### LINKED ISSUE #3325: LESU controller storage state bug
Everytime i re-enter my world the LESU controller storage state changes, for example, if my setup has 200M E if i leave my world and enter again it changes to 96 M E.

 - Minecraft: 1.21.1
 - Mod Loader: Fabric

ZurrTum:
When entering the game for the first time, the LESU network will depend on the order of block entities.

## Recurring bug:
The order of block entities depends on the hash order of pos objects.
net.minecraft.world.chunk.Chunk  Map<BlockPos, BlockEntity> blockEntities
blockEntities

## Question:
When storing block entity tick, the network will be found and joined only if the network is empty.
adding block entities to the network does not consider whether the surrounding areas have been checked, resulting in network isolation.
isolation
And because the order of LESU controller block entities is first, the machine will lose energy: 20M/20M E -> 4M/12M E
In addition: After destroying the middle one of the five LESUs, they are still recognized as four.
screenshots
You can even use a wrench to continuously refresh new capacity.
wrench

## Code refactoring goals:
In order to avoid sequence problems, it needs to be initialized in the onLoad event of the LESU controller. 
Because the findAndJoinNetwork method of LESUstorage is not used, surrounding squares are no longer skipped and all directions need to be tried, so optimization is necessary:
1. Record the connectable blocks around the block to greatly reduce the number of invalid visits to edge blocks
cut
2. Adding new blocks will directly expand without reconstructing the network
add
3. Keep track of the direction in which the block propagates, when deleting you only need to check the propagated blocks
change
4. Instead of checking the network every tick, let events drive it and update only when the user places blocks, breaks blocks, or uses a wrench.
5. Minimize the use of memory, use two bytes to save neighbors and links

## flaw:
1. The logic complexity of the code to delete the blocks increased, I used 240 lines of code.
2. Since my programming skills are not strong enough, there are some duplicate codes.
The code style made me struggle for a few days, and I added more comments later.
3. Although I tested adding, deleting, and multiple branches locally and everything was normal, it may still need to be tested by multiple people.


ZurrTum:
I've had time to make changes recently, please make suggestions if it doesn't meet the requirements.😄

modmuss50 (approved):
This looks great, nice work figuring it out and finding a solution. I havent done an indepth review but it all looks fine at a glance, a big improvement 👍

ZurrTum:
Thanks

PR #3333, opened 2024-11-26
 fix `ServerPlayerEntity` accessor breaking with alternate runtime mappings  

eristhea:
since https://github.com/TechReborn/TechReborn/commit/193be50df8dfa5d6294c3b1a93dbe5cad1895a36, TR has been incompatible with development workspaces using alternate runtime mappings (such as a Quilt dev workspace using [hashed](https://github.com/QuiltMC/mappings-hasher)), because of raw intermediary names being used without applying Fabric Loader's mapping resolver API to translate. this PR implements the proper usage of the floader API to fix a crash whenever a TR screen was opened!

note that that issue has been present since that commit (tech reborn `5.4.0`), so i'd really appreciate it if the fix could be backported to versions on `1.20` as well! the patch will apply to any version without conflict, so it shouldn't take too long :)

### DIFF
--- a/RebornCore/src/main/java/reborncore/common/screen/ServerPlayerEntityScreenHandlerHelper.java
+++ b/RebornCore/src/main/java/reborncore/common/screen/ServerPlayerEntityScreenHandlerHelper.java
@@ -24,6 +24,7 @@
 
 package reborncore.common.screen;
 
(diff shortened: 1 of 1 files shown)


modmuss50:
No, it works on Fabric we don't support any other platform.

Issue #3301, opened 2024-10-09
Output slot is not an option


ryankarch:
When clicking configure slots, and choosing a slot, it says "orange = output, blue = input, green = both" -- however, when clicking on a slot and toggling through the options, only blue and green show up, with no orange.

ryankarch:
Recreation:
https://github.com/user-attachments/assets/7f296433-70df-43a9-8310-02049d247337

PR #3334, opened 2024-11-26
Fix slot configuration colors

ZurrTum:
> Orange side means output.
> Blue side means input.
> Green side means both.

## Use old version colors:
https://github.com/TechReborn/TechReborn/blob/7528742fc0ba8e6ed56320d36656c694faa311c0/RebornCore/src/client/java/reborncore/client/gui/config/elements/FluidConfigPopupElement.java#L81-L86

## Fix
color
Maybe #3301 can be closed


modmuss50 (approved):
Good catch, this was a mistake.

Issue #3311, opened 2024-10-30
[Feature Request] Make slots configurable output order or maintain output order


ZurrTum:
### what is it:
![slots](https://github.com/user-attachments/assets/84280771-289a-450e-a1e2-d4c750146413)
**What is the output order?** 🤣

### Why is it needed?
I need to sort the items using storage units, but looking at the TechReborn code I see that the slot output is handled via a HashMap and the order changes every time I launch the game.

If there is a better way please tell me

> How I use storage units for sorting:
> 1. use all directions from the output to get the lowest priority direction (eg: North)
> 2. use storage units to pass items in that direction (eg: South to North)
> 3. If storage units are placed around a route (eg: in addition to South and North), items can be collected as the route passes
![in game](https://github.com/user-attachments/assets/552daf0b-3713-4eb9-914b-5775e06a9786)
![slot](https://github.com/user-attachments/assets/b2374e82-e181-49b4-8774-b24de1b401e9)
> Warning: unstable because it is implemented based on HashMap. The hashCode used is different every time you enter the game, causing the priority to change, and then a certain storage unit will be skipped.
> 4. Principle: By trying each direction until failure becomes the Low priority direction

### View Source Code:
**Use the minecraft Direction when creating:**
https://github.com/TechReborn/TechReborn/blob/a793dc92cbaae39ab6a68cac61acb3d9c185b97e/RebornCore/src/main/java/reborncore/common/blockentity/SlotConfiguration.java#L169-L173

**But HashMap uses the hashCode function return value of the minecraft Direction object, which will change, resulting in inconsistent ordering.**
https://github.com/TechReborn/TechReborn/blob/a793dc92cbaae39ab6a68cac61acb3d9c185b97e/RebornCore/src/main/java/reborncore/common/blockentity/SlotConfiguration.java#L192-L215

1. A simple idea is to use a **LinkedHashMap** implementation instead of a **HashMap**, so that the sorting maintains the insertion order, which is the enumeration order using Minecraft Direction. But it will maintain an extra list and I'm not sure how much of an impact it will have on performance
2. Another idea is to add reverse output option under "Auto Output" and use LinkedHashMap in reverse order
3. It would be better if you could set its priority via numeric keypad, but I feel the UI is too complicated

### Diff:
I fixed the output order to the order of minecraft Direction Enum through my own local modification.
![diff](https://github.com/user-attachments/assets/0eae1af3-c5a1-4a87-aade-ec4001159f7e)

[...]

PR #3335, opened 2024-11-26
Feature: support slot priority and pencil UI

ZurrTum:
screenshots
 ## UI features
- Added support for holding down the mouse to select multiple
- Below you can choose pencil to doodle directly

## Slot priority
- Support first or last
More see #3311


ZurrTum:
I didn't know if anyone needed it, so it took me a while to submit it 😄

modmuss50:
This looks really neat, im happy to merge it as is.

Not for this PR, but this whole interface has always been a bit confusing for users, and im not sure this change makes it any easier. I dont think there are many easy ways that this could be improved, but its worth thinking about.

PR #3346, opened 2024-11-30
[WIP] Port/1.21.4

ZurrTum:


Ayutac:
We don't use reflection to access private fields, that's what Access Wideners are for

ZurrTum:
mixin ?

Ayutac:
Nah, AccessWideners. Give me a minute.

Ayutac:
See https://github.com/TechReborn/TechReborn/commit/af48d9462b45cdbd9b72068c03b406f570e67fad (note that the AW of TR is generated automatically from the AW of Reborn Core)

PR #3337, opened 2024-11-28
Port to 1.21.4

modmuss50:
#### TODO

- [x] Compile
- [x] Run
- [ ] Work

PR #3355, opened 2024-12-03
Port/1.21.4 Add items folder to set up the model

ZurrTum:


modmuss50:
How did you create these? Ideally they should be datagenerated (The long term goal is to move all jsons to datagen).

Im happy to merge this to keep things moving but if possible we shouldnt be adding new jsons.

ZurrTum:
see #3337

modmuss50:
Ill merge this now as its an improvement, datagen might be much more work that can be done later.

PR #3362, opened 2024-12-05
[WIP] Port to 1.21.4

ZurrTum:
Synchronized recipe output since only the server now knows all recipes


ZurrTum:
In order not to repeatedly create pull requests, please wait for my work.

PR #3400, opened 2025-02-02
Compat PAL. Closes #3186

### LINKED ISSUE #3186: Quantum armor flight bug on servers only seems like

ZurrTum:
Use mixin to achieve compatibility, no need to include PAL jar package.
Dependence is only used in Mixin.

### DIFF
--- a/src/main/java/techreborn/compat/pal/MixinSimpleAbilityTracker.java
+++ b/src/main/java/techreborn/compat/pal/MixinSimpleAbilityTracker.java
@@ -53,17 +53,18 @@ public abstract class MixinSimpleAbilityTracker {
 		isAbilityAllowFlying = ability == VanillaAbilities.ALLOW_FLYING;
 	}
 
-	@ModifyReturnValue(method = "shouldBeEnabled()Z", at = @At("TAIL"))
+	@ModifyReturnValue(method = "shouldBeEnabled()Z", at = @At("RETURN"))
 	boolean checkAllowFlying(boolean enabled) {
 		if (isAbilityAllowFlying) {
 			if (enabled) {
 				TRAbility.allowFreeFlying(player);
+				return true;
 			} else {
 				TRAbility.dontAllowFreeFlying(player);
+				return TRAbility.ALLOW_FLYING.containsKey(player);
 			}
-			return TRAbility.ALLOW_FLYING.containsKey(player);
 		} else {
-			return false;
+			return enabled;
 		}
 	}
 
--- a/src/main/java/techreborn/compat/pal/MixinSimpleAbilityTracker.java
+++ b/src/main/java/techreborn/compat/pal/MixinSimpleAbilityTracker.java
@@ -1,0 +1,89 @@
+/*
+ * This file is part of TechReborn, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) 2025 TechReborn
+ *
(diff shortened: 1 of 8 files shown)


src/main/java/techreborn/compat/pal/MixinSimpleAbilityTracker.java
```diff
@@ -0,0 +1,89 @@
...
+import org.spongepowered.asm.mixin.Final;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.asm.mixin.Unique;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+import techreborn.init.TRAbility;
+
+@Mixin(SimpleAbilityTracker.class)
```
modmuss50, line 42: I dont think we should be mixining into other mods, there is no guarentee that the other mod will be stable.

modmuss50 (commented):
Im not keen on this at all, we should look into adding whats needed to Fabric API.

ZurrTum:
When there is no PAL, mixing not effective:
>[18:05:35] [main/WARN] (FabricLoader/Mixin) Error loading class: io/github/ladysnake/pal/SimpleAbilityTracker (java.lang.ClassNotFoundException: io/github/ladysnake/pal/SimpleAbilityTracker)
[18:05:35] [main/WARN] (FabricLoader/Mixin) @Mixin target io.github.ladysnake.pal.SimpleAbilityTracker was not found techreborn.pal.mixins.json:MixinSimpleAbilityTracker from mod techreborn
[18:05:35] [main/WARN] (FabricLoader/Mixin) Error loading class: io/github/ladysnake/pal/impl/VanillaAbilityTracker (java.lang.ClassNotFoundException: io/github/ladysnake/pal/impl/VanillaAbilityTracker)
[18:05:35] [main/WARN] (FabricLoader/Mixin) @Mixin target io.github.ladysnake.pal.impl.VanillaAbilityTracker was not found techreborn.pal.mixins.json:MixinVanillaAbilityTracker from mod techreborn

After mixing, it depends on the implementation of PAL, but at least better than conflict?
I don't know if it should be merged, so I submitted a pull request alone.

modmuss50:
This should be possible to do without using mixins, I would base it on the old PR here: https://github.com/TechReborn/TechReborn/pull/3241/files

I can maybe have a play with it if you want.

modmuss50:
Tottaly untested, but here is how I would do it without using a mixin: https://github.com/TechReborn/TechReborn/pull/3401

PR #3401, opened 2025-02-03
Optionally use PlayerAbilityLib for quantum flight

modmuss50:
Untested.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -20,6 +20,13 @@ repositories {
             includeGroup "dev.architectury"
         }
 	}
+    maven {
+        name "Ladysnake Mods"
+        url "https://maven.ladysnake.org/releases"
+        content {
+            includeGroup "io.github.ladysnake"
+        }
+    }
 	mavenCentral()
 }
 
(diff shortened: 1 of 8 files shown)


modmuss50:
Im really not a fan of this, it should have been put into fabric api years ago. Doing this will likely mean that will never happen.

ZurrTum:
Thank you, I did not expect this way of implementation.
I don't know how many people need it, I just take this as a challenge.

DreiPixel:
Will crash Singleplayer if there is no !world.isClient check.
the mod "Origins" has PlayerAbilityLib included even on Singleplayer. 
so if you have techreborn (with that patch) and origins, using the Quantum armor will crash the game.

https://github.com/Ladysnake/PlayerAbilityLib/blob/4da7e89828ffae97203d68c1dd8766111d431cc5/src/main/java/io/github/ladysnake/pal/impl/PlayerAbilityView.java#L46

ZurrTum:
This modification is now integrated into version 1.21.1 and has been tested and used normally with PAL items.

Issue #2402, opened 2021-05-27
Upgrades for Auto Crafting Tables


Snoe0:
**Is your feature request related to a problem? Please describe.**
Auto Crafting tables were fixed in the recent beta version, but they are extremely slow and do not allow for upgrades, so adding more of them is the only way to make it faster.

**Describe the solution you'd like**
Allow for upgrades like the Overclock Upgrade usable in the Autocrafter (give it the 4 slots like other machines).

**Describe alternatives you've considered**
Make an upgraded version that is faster.

**Additional context**
![image](https://user-images.githubusercontent.com/83789327/119752275-b48ecb80-be6a-11eb-9f3d-fbd8e5a3db99.png)

PR #3407, opened 2025-02-23
Make Auto Crafting Table upgradable

Dabolus:
While trying to automatize some processes in the game, I noticed that the Auto Crafting Table doesn't currently allow upgrading. This creates a bottleneck in automation processes, so I tried checking if there were some open issues here in the repo.

I found [this very old issue](https://github.com/TechReborn/TechReborn/issues/2402) that has been closed in favor of [this discussion](https://github.com/TechReborn/TechReborn/discussions/2533), but neither of them has had any activity.

For this reason, I decided to try adding support for upgrades by myself, based on how upgrades are handled on other machines. I added and checked support for all the upgrades, including muffler. Let me know if I'm missing something else. You can find some screenshots below.

**Upgrades UI**

image

**Energy storage + transformer upgrades**

image

**Overclocker upgrades**

Screen Recording 2025-02-23 at 15 49 57


ZurrTum:
Thank you. I adjusted it according to the following list:
1. Use static final value
2. Use canBeUpgraded default function
3. Updated only when progress is 0
4. Limit the sound cycle to recipe time

Dabolus:
Thank you! 🙏🏻

PR #3425, opened 2025-05-09
Break out of iteration after setting iterator to `null`

burdoto:
This change is necessary because otherwise, in case the loop would continue, the game will [crash](https://mpaste.ntms.link/paste/?content=https%3A%2F%2Fcdn.discordapp.com%2Fattachments%2F1281630168552702022%2F1369899409562931220%2Fcrash-2025-04-29_07.43.29-server.txt&raw=true) because the iteration target of the `foreach` loop is now null.


burdoto:
actually, after a moment of consideration, i realized this change is not cause of the problem because the iterator has already been obtained at that point;
rather for some reason `targets` remains to be `null` even after the preceding block that checks for exactly that.

Issue #3427, opened 2025-05-11
Player Detector range measured from corner of block not center


James-Schaffer:
**Describe the bug**
Player detector measures from its coordinate which in world space is the bottom corner. Means that at 1 block radius (although similar issues arise with any radius) standing on one side of the block will detect you whilst the other side will not.

**Steps to Reproduce**
Place player detector in 1.21.1 and test redstone signal going out at low radius values

**Environment (please complete the following information with the version):**
 - Minecraft: 1.21.1
 - Mod Loader: fabric

James-Schaffer:
on PlayerDetectorBlockEntity.java, line 97

if (MathHelper.sqrt((float)player.squaredDistanceTo(pos.getX() +0.5f, pos.getY() +0.5f, pos.getZ() +0.5f)) <= (float)radius ) {



suggested fix ^

PR #3428, opened 2025-05-11
Proposed fix for : Player Detector range measured from corner of block not center #3427

James-Schaffer:
Fix for bug :

Player Detector range measured from corner of block not center #3427

Added offset so distance calculation taken from center of block not the corner


ZurrTum:
Thank you

PR #3481, opened 2026-02-03
Add recipes needed for the Nuclear Reactor

Dabolus:
Opening this PR to gather some feedback since this involves a lot of moving pieces and the flow to craft the reactor and its fuel is different from other mods.

Note: all the textures in this PR (besides the existing ones) were made by me, so I'm also open to feedback on that side.

---

First of all, uranium has been added to the ore gen. Below you can see uranium ore, deepslate uranium ore, raw uranium, and all the related storage blocks:

This is how they look like in the world:

---

As with other raw blocks, uranium ores can be grinded in the grinder to get raw uranium:

image

---

Raw uranium then needs to be grinded in the industrial grinder either with water or with sodium persulfate (to increase the output) to get uranium dust:

image

image

---

Uranium dust is composed mainly by uranium-238 and a tad bit of uranium-235. It needs to be enriched to be used as nuclear fuel. To enrich uranium, you must gasify it first. This is done by converting it to uranium hexafluoride by making it react with fluorine in the chemical reactor:

image

---

Fluorine can be obtained either by electrolyzing granite, or by electrolyzing bone meal with sulfuric acid:

image

---

Uranium hexafluoride must then be centrifuged to separate uranium-238 from uranium-235. The fluorine will be lost in the process and replaced with compressed air:

image

---

To produce the fuel pellets, uranium needs to be enriched and sintered. This is done by placing uranium-238 together with more uranium-235 in the blast furnace, which will produce uranium fuel pellets:

image

---

Finally, the fuel pellets need to be inserted into empty fuel rods using the solid canning machine:

image

---

The empty fuel rods are obtained by rolling 3 magnalium plates in the rolling machine (to simulate magnox fuel rods):

image

---

Uranium fuel rods can also be combined in dual/quad fuel rods:

image

---

Now that we have the fuel, we need the nuclear reactor. The nuclear reactor is made out of a very strong alloy, as it needs to resist high temperatures and corrosion. This material, called industrial alloy, needs to be crafted by combining multiple different metals into an industrial alloy ingot first:

image

The materials are: steel plates, invar ingots, chromium ingots, and a small pile of manganese dust.

---

The industrial alloy ingots then need to be pressed into industrial alloy plates using the implosion compressor:

image

---

These plates can be crafted together with an industrial machine frame to get a reactor chamber:

image

---

Finally, 3 of these reactor chambers are used, together with 4 more plates and a couple more materials, to craft the nuclear reactor:

image

Up to 6 more reactor chambers can be added around the nuclear reactor to increase the number of its internal slots.

---

The PR also includes the recipes for all the other components (vents, exchangers, platings) for which I'm not adding the screenshots below as this description is already quite long.


Honzapkcz:
Amazingly done.

Note that you can get 8x uranium 238 out of one hexafluorite cell which is made from only one uranium 238 dust.
[centrifuge recipe here](https://github.com/TechReborn/TechReborn/pull/3481/changes/133058b64e5d2dccf1736fbaf759661d0b60a055#diff-19558a492407903eaa82951d51b603860614f5504acbc267b22fa8c32f6dcb31)

I would also add some UU-Matter crafting recipes for the ore.

Also, why not add raw lead as a byproduct of industrial grinding with sodium persulfate?
Citing wikipedia's [page](https://en.wikipedia.org/wiki/Uraninite):
> [Radioactive decay](https://en.wikipedia.org/wiki/Radioactive_decay) of the uranium causes the mineral to contain [oxides](https://en.wikipedia.org/wiki/Oxide) of [lead](https://en.wikipedia.org/wiki/Lead) and trace amounts of [helium](https://en.wikipedia.org/wiki/Helium).

Dabolus:
Thanks for the feedback!

---

> Note that you can get 8x uranium 238 out of one hexafluorite cell which is made from only one uranium 238 dust.

It's probably not very clear from the screenshots as the texture is very similar, but the dust used to make the Uranium Hexafluoride cell is the raw **Uranium Dust**, not **Uranium-238 Dust**. From a single Uranium Dust, you can get (via the Uranium Hexafluoride) 8x Uranium-238 Dust and 1x Uranium-235 Small Dust. These two can then be either combined with a higher percentage of Uranium-235 to enrich the Uranuim, or recombined in the same percentages on the Crafting Table to get the raw Uranium Dust back, so duplication shouldn't be possible.

image

---

> I would also add some UU-Matter crafting recipes for the ore.

I wondered why old tech mods don't have it and I found [this explanation](https://www.reddit.com/r/feedthebeast/comments/14dubg/comment/c7c6o28/) that makes sense. However, considering everything that's needed to make use of the uranium later on, I think it wouldn't be an issue to add it. I will check which recipes are still free and try to add something.

---

> Also, why not add raw lead as a byproduct of industrial grinding with sodium persulfate?

That's a very good idea! I'll add it. 🙏🏻

modmuss50 (approved):

PR #3487, opened 2026-04-09
Implement custom fluid ingredient

Dabolus:


RebornCore/src/main/java/reborncore/common/fluid/container/FluidContainerIngredient.java
```diff
@@ -0,0 +1,192 @@
...
+
+	private static final long DROPLETS_PER_MB = FluidConstants.BUCKET / 1000;
+
+	private final Holder<Fluid> fluid;
+	private final long amountMb;
+
+	public FluidContainerIngredient(Holder<Fluid> fluid, long amountMb) {
+		this.fluid = fluid;
+		this.amountMb = amountMb;
+	}
```
modmuss50: Nit pick: Could maybe be a record?

modmuss50 (approved):

RebornCore/src/main/java/reborncore/common/fluid/container/FluidContainerIngredient.java
Dabolus: Yes, good catch ✅

PR #3508, opened 2026-06-02
Fixed PR #3506 to also include updating nearby cables of the chambers

Xurceo:
- Registered side energy storage in ModRegistry.java
- Added functionality for reactor chambers to push energy to nearby blocks
- Made so that when reactor placed or removed chambers will connect or disconnect from cables


modmuss50 (approved):
Looks good thanks.

PR #3514, opened 2026-06-04
Allow users to edit cable transfer rates in config

Swagi-REDACTED:
also expanded battery blocks which appear in machines.json!


Swagi-REDACTED:
Tested on 26.1.2 worked perfectly!

modmuss50 (approved):

PR #3533, opened 2026-07-30
feat: REI compat

TonimatasDEV:
This PR ports the 1.21.10 REI compat code to 26.2.

I added Architectury and Cloth Config because it is not resolved automatically.


modmuss50:
Looks good thanks, it wasnt clear if REI was going to be updated or not. I think we can try to maintain support for both.

modmuss50 (approved):

PR #3535, opened 2026-08-03
Copper nugget update

Furikorifuto:
.tech reborn copper nugget to vanilla copper nugget
.Heaby core recipe fix


maciejstepie:
seems like you forgot to sync with the branch repo after adding cinnabar

PR #3536, opened 2026-08-03
copper update

Furikorifuto:
.TR copper nugget to MC copper nugget.
.heavy core recipe fix
.scrap box code cleanup


RebornCore/run/logs/2026-08-03-2.log.gz
modmuss50, line 1: Why are these files here?

PR #3538, opened 2026-08-04
ore loot and blast furnace

Furikorifuto:
.bauxite galena sheldonite ore drop dust.
.blast furnace copper gear recipe


maciejstepie:
Thats a lot for chromium and titannium heat in blast furnance. Its needed to upgrade it to get that heat with Industrial Machine Casing

Furikorifuto:
I also thought we might need the steel casing process.
If there are a lot of objections, I'll revert the blast furnace recipe.
basic:1860Heat
advanced:2812Heat
industrial:3730Heat

maciejstepie:
Yeah, stell could go up a bit to 1,5k i guess. Just not sure about the titanum.  It may slow a lot of progress since it made for advanced tools.

Furikorifuto:
I set the titanium back to 1500.

maciejstepie:
Don't get me wrong. It think titanium should require more heat, 2500 is too much in my opinion

PR #3542, opened 2026-08-13
Recipe rebalance

Furikorifuto:
.Some recipes to data gen
.Added an aluminum recipe
(cell and basic machine frame)
.Nerf to the number of glass fiber cables
.The three types of machine casings recipe rebalance.


Furikorifuto:
I'll adjust it again

